PR #5, opened 2024-06-28
Add Awesome Low-Fat JS

terrablue:


patak-dev:
Hey! Thanks for the PR, for now, we'll wait a bit to see how the your awesome list evolves. We'll also may maintain a curation here, so it may not be the best fit. Good luck with the initiative though, thanks for sharing it!

PR #13, opened 2024-07-13
docs(cleanup): add recommendation to use package-size-calculator

TheDevMinerTV:

PR #14, opened 2024-07-13
docs(cleanup): add recommendation to use package-size-calculator

TheDevMinerTV:
Supersedes #13 (idk, github just closed my PR for some reason)

### DIFF
--- a/docs/guide/cleanup.md
+++ b/docs/guide/cleanup.md
@@ -82,6 +82,8 @@ Even if you will try tackle them yourself, this will help improve visibility so 
 
 Or if you do not have time to tackle it yourself, it is possible someone else can follow your findings.
 
+Feel free to also attach a package report using [DevMiner's package-size-calculator](https://github.com/TheDevMinerTV/package-size-calculator) to the issue, to give an overview over the package's size and the changes you are proposing.
+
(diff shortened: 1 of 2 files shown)


docs/guide/cleanup.md
```diff
@@ -82,6 +82,9 @@ Even if you will try tackle them yourself, this will help improve visibility so
 
 Or if you do not have time to tackle it yourself, it is possible someone else can follow your findings.
 
+Feel free to also attach a package report using [DevMiner's package-size-calculator](https://github.com/TheDevMinerTV/package-size-calculator) to the issue, to give an overview over the package's size and the changes you are proposing.
+![Package Report](../public/pkgcli-size-report.png)
```
patak-dev, line 86: Nice! I think we should leave only the paragraph. If we'd like to add the image, we should have it as a code block but probably in this case it would be a bit much for the current state of the guide. You can also include this tool in the Resources guide.

TheDevMinerTV: I've removed the screenshot and added the tool to the resources page as well.

patak-dev (approved):

bluwy (approved):

PR #16, opened 2024-07-25
Correct spelling, grammar, punctuation

DePasqualeOrg:

PR #23, opened 2024-08-06
Improve grammar across the site

bvisness:
This incorporates many of the changes that were abandoned in #16, plus a few other scattered changes.

As I mentioned on Discord, I am an Oxford comma enthusiast, and believe they should always be used in all situations. If we really would prefer to omit the Oxford comma across the site, I will begrudgingly update my PR to do so, but not without complaining :stuck_out_tongue: 

As for the perhaps contentious issue of "cleanup" vs. "clean up" - in most cases, I have gone with "cleanup", with the understanding that it is to be thought of in context like "the e18e cleanup initiative". I have in one or two places phrased it "clean up" where it is explicitly used as a verb (e.g. "clean up the ecosystem").

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,15 +17,15 @@
 
 The e18e (Ecosystem Performance) project is an initiative to bring together the groups and individuals who are passionate about improving performance of the JavaScript ecosystem.
 
-Many ongoing efforts are already happening in this space, from [dependency tree cleanups](https://github.com/43081j/ecosystem-cleanup) to [performance optimizations](https://marvinh.dev/blog/speeding-up-javascript-ecosystem/), and much more.
+Many ongoing efforts are already happening in this space, from [dependency tree cleanups](https://github.com/43081j/ecosystem-cleanup) to [performance optimizations](https://marvinh.dev/blog/speeding-up-javascript-ecosystem/) and much more.
 
-Our aim is to provide a space for contributions, ideas and knowledge sharing around the importance of performance across the ecosystem. Hopefully, with this collaboration, we can improve visibility of the impact of things like dependency choices. Join us at the [e18e Discord Server](https://chat.e18e.dev) to connect with other like-minded devs.
+Our aim is to provide a space for contributions, ideas and knowledge sharing around the importance of performance across the ecosystem. Hopefully, with this collaboration, we can improve visibility of the impact of things like dependency choices. Join us at the [e18e Discord server](https://chat.e18e.dev) to connect with other like-minded devs.
 
 Currently, there are three main areas of focus:
 
-- [cleanup](https://e18e.dev/guide/cleanup) - cleaning up dependency trees and modernizing popular tools and libraries across the ecosystem.
-- [speedup](https://e18e.dev/guide/speedup) - speeding up parts of the ecosystem many of us depend on.
-- [levelup](https://e18e.dev/guide/levelup) - documenting and providing modern, lighter alternatives to established tools and libraries we all regularly use.
+- [Cleanup](https://e18e.dev/guide/cleanup): cleaning up dependency trees and modernizing popular tools and libraries across the ecosystem.
+- [Speedup](https://e18e.dev/guide/speedup): speeding up parts of the ecosystem many of us depend on.
+- [Levelup](https://e18e.dev/guide/levelup): documenting and providing modern, lighter alternatives to established tools and libraries we all regularly use.
 
 ## Contributing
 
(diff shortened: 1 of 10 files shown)


docs/guide/cleanup.md
```diff
@@ -31,102 +31,92 @@ We're mostly looking for packages which fit the following criteria:
...
 
-For these dependencies, we should first find out if there's a reason we need them and not an alternative.
+For these dependencies, we should first find out if there's a reason the dependency is necessary and shouldn't be replaced with an alternative.
 
 Common things to check:
 
-- node `engines` specifying a node version low enough that this dependency supports (and alternatives don't).
-- there is no known alternative (good opportunity to make one!).
-- every deep dependency in the subtree is necessary (in many cases the tree may be deep but for good reason, to share modules, etc).
+- [`engines`](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines) specifying a lower Node version that this dependency supports, but alternatives don't
```
43081j, line 93: should this be `than this dependency supports`?

bvisness: No, but I realize the overall wording is confusing. I can take another crack at it.

43081j: it is 😅 

i forget what point it was trying to get across

we check `engines` for two reasons usually:

- if it is high enough that some dependencies are made redundant by native functionality
- if it is low enough that we can't yet move to a more modern alternative

i guess this was saying you should check if the constraint is low enough that alternatives can't be used? (the 2nd point)

docs/guide/speedup.md
```diff
@@ -35,12 +35,12 @@ The code you write plays an important role in the performance of your app. Some
...
 
 ### Avoid chaining array methods
 
 Chaining array methods like `map`, `filter`, `reduce`, etc leads to many intermediate arrays being created and disposed, causing more work for the garbage collector. Each chain also leads to an extra iteration, which can be more times than needed.
 
-Prefer using [`for` loops](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for), [`for...in` loops](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of), and[`for...of` loops](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of), or a single chain method only to prevent the caveats above.
+Prefer using [`for` loops](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for), [`for...in` loops](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of), and[`for...of` loops](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of), or a single chain method to prevent the above issues.
```
zettca, line 46: Unrelated to grammar, but a typo I found 👀 
```suggestion
Prefer using [`for` loops](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for), [`for...in` loops](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in), and[`for...of` loops](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of), or a single chain method to prevent the above issues.
```

docs/guide/cleanup.md
bvisness: I have rewritten this section to try and be clearer from the start.

README.md
```diff
@@ -17,15 +17,15 @@
...
 
 Currently, there are three main areas of focus:
 
-- [cleanup](https://e18e.dev/guide/cleanup) - cleaning up dependency trees and modernizing popular tools and libraries across the ecosystem.
-- [speedup](https://e18e.dev/guide/speedup) - speeding up parts of the ecosystem many of us depend on.
-- [levelup](https://e18e.dev/guide/levelup) - documenting and providing modern, lighter alternatives to established tools and libraries we all regularly use.
+- [Cleanup](https://e18e.dev/guide/cleanup): cleaning up dependency trees and modernizing popular tools and libraries across the ecosystem.
+- [Speedup](https://e18e.dev/guide/speedup): speeding up parts of the ecosystem many of us depend on.
+- [Levelup](https://e18e.dev/guide/levelup): documenting and providing modern, lighter alternatives to established tools and libraries we all regularly use.
```
43081j, line 28: should these ones be lowercase? as "brand" names?

43081j (approved):

43081j:
looks good to me

i think you need to catch up from `main`, though

i was also still unsure on `Speedup` etc. I would've though the choice is between `speedup` or `Speed up` (the former acting as a kind of brand name)

README.md
bluwy: I personally think it's ok. It's like webpack but people call it Webpack all the time, so might as well embrace it 😄

PR #29, opened 2024-09-10
docs: correct phrase about PostCSS in August blog post

patak-dev:
They are writing about reducing node_modules size since 2017 (!) 😎
https://evilmartians.com/chronicles/autoprefixer-7-browserslist-2-released
https://x.com/PostCSS/status/1276669602324504580


43081j:
Good catch! 🙏

PR #31, opened 2024-10-12
docs: add `pkg-graph` to websites

yuyinws:
Another website tools to visualizes a package's dependency graph.
It base on Webcontainer, so when analyzing large packages, it will be faster than other similar tools.

Homepage: https://pkg-graph.info/
GitHub: https://github.com/yuyinws/pkg-graph.info


bluwy:
> It base on Webcontainer, so when analyzing large packages, it will be faster than other similar tools.

How so? I use https://npmgraph.js.org more often, and it's much efficient by only downloading the packument (not the entire tarball and unpacking). For comparison (warm visit, resources from other runs cached):

- https://npmgraph.js.org/?q=astro:
  - Time to load graph: 8s
  - Network transfer size: 19MB
  - Uncompressed resource: 100MB
  - Request count: 300

- https://pkg-graph.info/astro
  - Time to load graph: 35s
  - Network transfer size: 37MB
  - Uncompressed resource: 183MB
  - Request count: 747

PR #35, opened 2024-11-06
docs: remove X links

patak-dev:
In just a few days, the [e18e's Bluesky account](https://bsky.app/profile/e18e.dev) is about to have the same number of followers than the X one.

I think we should focus our efforts there, and the X one is no longer a good tool to reach other folks.

This PR removes the X links and points people to the Bluesky profile instead (that is public and can be viewed without a Bluesky account). We should also remove the X link from the e18e GitHub org.


SuperchupuDev (approved):

TheDevMinerTV (approved):
dead platform 👏

PR #43, opened 2024-12-26
chore: bump node version on netlify

iiio2:
Since node version 22 is stable now, we may use node version 22 here.


iiio2:
Thanks @bluwy .

bluwy:
No problem. Thanks for the contribution!

PR #49, opened 2025-02-14
chore: simplify package manager in scripts

bluwy:
This one is a little subjective, but I figured to use pnpm directly instead of `@antfu/ni` since the user would've already needed pnpm to install the project


patak-dev (approved):

43081j:
seems fair to me

you will need to rebase FYI since your other PR got merged

PR #52, opened 2025-02-14
chore: delete CODE_OF_CONDUCT.md

bluwy:
We can remove the file here now that it's located at https://github.com/e18e/.github/blob/main/CODE_OF_CONDUCT.md, which applies to all repos under the org


43081j:
Can you do a quick search to see if there's anywhere we reference this one? 

Maybe in pinned discord messages etc

We should update them at the same time

bluwy:
Yeah, it looks like we have to update the link in the `#about` channel on discord. I don't see it mentioned anywhere else even in our website or socials.

patak-dev (approved):

PR #54, opened 2025-03-19
docs: add migration to esm post

43081j:


43081j:
@patak-dev could you help me with a cover image? 👀 

@joyeecheung could you double check that the `require(esm)` section isn't nonsense? and if you have a useful link of where most of it is explained, i could link to it

docs/blog/migrating-the-ecosystem-to-esm.md
```diff
@@ -0,0 +1,140 @@
...
+
+As part of this migration, we have three types of package to deal with:
+
+- ESM packages
+- CommonJS packages
+- Dual packages
+
+### Migrating a CommonJS package
+
+In most cases, this is as simple as setting `type` to `"module"` in `package.json`, then updating all imports to include file extensions.
```
ljharb, line 55: can this also mention the more performant approach of renaming all the files to ".mjs", and for a package, adding `exports` so that external consumers don't have to use file extensions?

docs/blog/migrating-the-ecosystem-to-esm.md
```diff
@@ -0,0 +1,140 @@
...
+```json
+{
+  "name": "my-package",
+  "type": "module"
+}
+```
+
+### Migrating a dual package
+
+Dual packages are basically a necessary evil if you want to support both CJS and ESM _before `require(esm)` was available_.
```
ljharb, line 68: this isn't quite accurate; a dual package isn't necessary to support both CJS and ESM because you can `import` CJS. The reasons for a dual package are far more nuanced, and don't actually exist for most packages that have shipped dual.

docs/blog/migrating-the-ecosystem-to-esm.md
43081j: that's true, i'll add an extra paragraph 👍

docs/blog/migrating-the-ecosystem-to-esm.md
ljharb: this is true only if they insist on writing native ESM, though - that's the missing nuance here, i think. iow, shipping CJS-only is a trivial way to avoid any need for a dual package (prior to `require(esm)`)

43081j: that's also not quite true. various packages ship dual so they can support browsers natively without the need for a bundler. they had to use ES modules because browsers don't understand CJS, so they had to ship dual to also support older Node (or require a dynamic import).

i think this shows there's too many edge cases and situations here to mention without going off on a huge tangent

ljharb: that's true.

maybe just say "support both CJS and non-node ESM"?

docs/blog/migrating-the-ecosystem-to-esm.md
```diff
@@ -0,0 +1,170 @@
...
+  "type": "module",
+  "tsup": {
+    "format": ["esm"]
+  }
+}
+```
+
+## How the community is helping
+
+Within the e18e community, we have been tracking this in an [issue](https://github.com/e18e/ecosystem-issues/issues/129) for some time.
```
patak-dev, line 125:
```suggestion
Within the e18e community, we have been tracking the migration from CJS to ESM in an [issue](https://github.com/e18e/ecosystem-issues/issues/129) for some time.
```

docs/blog/migrating-the-ecosystem-to-esm.md
```diff
@@ -0,0 +1,175 @@
...
+```
+
+And updating the imports:
+
+```ts
+// before
+import './foo'
+
+// after
+import './foo.js'
```
joyeecheung, line 87: Since the title says

> Migrating a CommonJS package

Then I assume `foo.js` is a CJS module? I think in that case, it must be either renamed to `foo.cjs` if `"type": "module"`, or the content of `foo.js` itself must be migrated to be authored in ESM.

If `foo.js` still contains CJS code, I am not sure if this really counts as migration - it just adds an extra ESM wrapper around the code authored in CJS, which isn't necessarily in the first place because `import cjs` already works, and this is just `import esm-wrapper` where `esm-wrapper` imports `cjs` again?

If it means `foo.js` must also be rewritten to be in ESM format, then from my reading this is missing a sentence to clarify that "first, rewrite `foo.js` to be in ESM format".

docs/blog/migrating-the-ecosystem-to-esm.md
```diff
@@ -0,0 +1,175 @@
...
+
+These packages work by having **two copies** of the sources and the types.
+
+Fortunately, this doesn't increase the package size much since it will be compressed anyway. However, **it does increase the on-disk size** (once extracted by `npm`).
+
+Most people shipping TypeScript packages like this will be using a tool like [tsup](https://github.com/egoist/tsup) or [tshy](https://github.com/isaacs/tshy).
+
+Those shipping JavaScript often just use a bundler like [esbuild](https://github.com/evanw/esbuild) to create two bundles (or one and the sources).
+
+To migrate from these setups, we mostly need to do the same steps as migrating a CommonJS package from above.
```
joyeecheung, line 112: Not sure how `tsup` actually manages the exports, but from my understanding, usually migrating a dual module means adding `module-sync` to the exports: https://github.com/nodejs/node/pull/54648

So for example, if the exports map originally looks like this (it uses `node` to provide additional fallback for non-node environments, like what babel is doing):

```js
{
  "type": "module",
  "exports": {
    // On Node.js, provide a CJS version of the package transpiled from the original
    // ESM version, so that both the ESM and the CJS consumers in the same graph get
    // the same version to avoid the having two versions of the same package
    // conflicting with each other a.k.a. package hazard.
    "node": "./dist/index.cjs",
    // On any other environment, use the ESM version.
    "default": "./index.js"
  }
}
```

To prepare to drop dual shipping (say v1 is still dual-shipping because it is supporting older versions of Node.js, but it still wishes to tell Node.js > 20 to always pick the ESM version, no matter it's required or imported)

```js
{
  "version": "1.4.0",
  "type": "module",
  "exports": {
    "node": { // Packages can drop this special case as they drop support for older Node.js
      // On new version of Node.js, both require() and import get the ESM version
      "module-sync": "./index.js",
      // Supply ESM to bundlers for better generated code
      "module": "./index.js",
      // On older version of Node.js, where "module" and require(esm) are not supported,
      // use the transpiled CJS version to avoid dual-module hazard.
      "default": "./dist/index.cjs"
    },
    // On any other environment, use the ESM version.
    "default": "./index.js"
  }
}
```

To actually drop dual shipping (say on v2 which only supports Node.js v20.x and above):

```js
{
  "version": "2.0.0",
  "type": "module",
  "exports": {
    ".": "./index.js"
  }
}
```

There are more complex cases where people define both `require` and `import` exports, which needs extra wrapping to avoid the dual package hazard. But in those cases the migration (without bumping major to avoid changing Node.js support matrix) would still boil down to: add a `module-sync` exports (which must preceed `require`, if there is one) that provides the ESM exports, so that both require and import picks the ESM version. And when it bumps major to drop support for Node.js <20, simply remove all these mess and just point the exports to the ESM, without caring how it's loaded.

joyeecheung (commented):
Sorry for the late reply, was swamped by meetings this week...

docs/blog/migrating-the-ecosystem-to-esm.md
43081j: It is an es module

This snippet is assuming you already converted your syntax to esm

docs/blog/migrating-the-ecosystem-to-esm.md
43081j: This is written under the assumption you just go all in on esm. Which presumably means you have one very simple export (your last example), which works in most node versions as esm but only 20.x and above as cjs

joyeecheung: I think `module-sync` primarily concerns whether the package is still supporting Node.js < 20. If they already decide to bump major and drop support for Node.js < 20 then yes they do not need to care about `"module-sync"`. Maybe a note about Node.js version support would help clarifying that.

docs/blog/migrating-the-ecosystem-to-esm.md
joyeecheung: That makes sense now. Not sure if it's just me not reading thoroughly enough, it does feel that this assumption is missing somewhere...

docs/blog/migrating-the-ecosystem-to-esm.md
43081j: node < 20 would still support it without `module-sync` would it not? as an ES module

it just means you can't `require(esm)` it. which is fine in this all-or-nothing strategy

joyeecheung: > node < 20 would still support it without module-sync would it not? as an ES module

Yes, it can be imported, just not required. Though for packages that adhere to the "when dropping support for older Node.js versions, bump major" principle, it affects what they could do with the last major release before the migration. For example:

- The package is dual-shipped in v1, which supports Node.js 18.
- The package wants to become ESM-only, so it just discards the CJS distribution, ditch the transpilation step, and ship in ESM form as-is, this is now v2 of the package, since it will no longer be require-able on Node.js 18.

The article has covered the two points above, though there is a remaining question:

- Is the package still maintaining v1 for Node.js > 20 for some time, or are they completely abandoning it and moving on to v2?

If they are still maintaining v1 (e.g. if v2 is not just about shipping ESM - which is likely not observable for users - but also has some other significant and observable major changes, which means users can still cling on to v1 for a while), then `"module-sync"` allows them to specify that on Node.js > 20 is always loading the ESM version to reduce dual-package hazard on v1. Or, even if they are abandoning v1, this addition of `"module-sync"`  might be the last thing they'll release to v1 to reduce the issues that may come from the dual-package hazard.

43081j: That makes sense. In the packages that migrated so far as part of this coordinated effort, they did all just release new majors which contained nothing else. Just a switch to esm only

So none needed module-sync since it was generally understood that anyone who wants to require it from now on needs node 20 and above

joyeecheung: Oh I see, is this article just about "what have been done", instead of "what should be done if you are going to migrate to ESM"? If it's the former, then it makes sense 👍

PR #58, opened 2025-04-13
docs: link to the main org instead of the website

TheAlexLichter:
To get more info around e18e, linking to the GitHub organization and their README should be better than to this repo ☺️


43081j (approved):
makes sense to me 👍

PR #60, opened 2025-04-17
docs: improve wording and make content more direct

outslept:
English is not my first language, so I would greatly appreciate if someone could review these changes to ensure everything is properly phrased and clear to understand. Please suggest any further improvements if needed. Also we can catch up the changes from here https://github.com/e18e/e18e/pull/23

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -7,7 +7,7 @@ titleTemplate: Ecosystem Performance
 hero:
   name: e18e
   text: Ecosystem Performance
-  tagline: Cleanup, Speedup, Levelup.<br>One Package at a time.
+  tagline: Cleanup, Speedup, Levelup.<br>One package at a time.
   image:
     src: /logo.svg
     alt: e18e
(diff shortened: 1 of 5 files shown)


docs/index.md
```diff
@@ -22,11 +22,11 @@ hero:
 features:
   - icon: ✨
     title: Cleanup
-    details: Improving packages by removing redundant dependencies or replacing them with modern alternatives.
+    details: Removing redundant dependencies and replacing them with modern alternatives.
     link: /guide/cleanup
   - icon: ⚡️
     title: Speedup
-    details: Improving performance of widely used packages.
+    details: Enhancing the performance of widely used packages.
```
43081j, line 29: maybe its just me, but i feel like `Improving` sounded better before 🤔

43081j (approved):

docs/index.md
outslept: It was Improving, Improving, Building before, so I wanted to be more "playful" to not repeat ourselves there

What do you think about some alternatives for this?

43081j: i think this one would be fine as it was but with `the` at least

> Improving the performance of widely used packages

i think a small amount of repetition is fine, as improvement makes more sense here than enhancement

outslept: Oki, resolved with latest commit 👀

PR #67, opened 2025-05-20
feat: add Advocacy guide page

### LINKED ISSUE #66: Add a guide for how to suggest using a new package
👋 I maintain a bunch of packages that have had e18e package swaps suggested on them. Although I'm personally in favor of the general initiative _(faster performance! smaller installs!)_, not everyone in open source is aware of it. A lot of these issues are being posted without much context or any metrics justifying the changes. This is a problem because it makes it seem like the changes are arbitrary or personal-preference-based.

Proposal: I'd like to write a guide for the e18e site to how to suggest using a new package in an existing repository. Most importantly, I'd like to call out how to justify the common points of improvement e18e-promoted packages come with:

* **Bundle size** (for built/bundled code): showing the actual difference in the bundle size, not just the dependency tree
  * When a direct dependency is also a deep dependency, swapping to a new direct dependency can result in a _bigger_ bundle size
* **Dependency tree**: showing the actual different in dependency tree, not just that the new package has a smaller tree on its own
  * Same reasoning with bundle size
* **Speed**: posting performance measurements of the _specific package_, not general benchmarks
  * Isolated benchmarks aren't very useful; downstream users often have very different usage scenarios and/or don't use some dependencies in hot paths

I also think it'd be useful to mention what _isn't_ convincing for many maintainers:

* Abstract bug reports / feature requests: indicating that a new dependency could offer bug fixes, improved DX, more features, etc. in the abstract is not convincing without any evidence those impact the project
  * For bug fixes, be sure to have a reproduction -- much like any other kind of bug report
  * For DX and features, you'll similarly want to be able to _show_ why this is better
* npm download stats: many maintainers don't see this as a reliable metric; saying a package is trustworthy because of it -even alongside other, better evidence- can actually _weaken_ the point
  * Download stats can be easily gamed, such as a package gaining many downloads from other packages by the same owner (not maliciously, just coincidentally)
  * If you're trying to justify "trust" of a package, pointing to large existing consumers having switched over is generally more convincing
* Overly expansive support ranges: an existing dependency supports a much wider version than necessary is not an issue on its own; it's better to point to the _impact_ of that wide range 
  * Example: if an existing dependency has a lot of code to support old versions of Node.js, show that the replacement is faster / a smaller size / etc. as the evidence

I chatted a bit with @43081j about this. General questions:

* Does such a thing already exist? (I had a memory of seeing a PR somewhere, but can't find it now...)
* Does this sound reasonable as a general narrative, before I start writing up a whole thing?
* Are there other points that should be added?

JoshuaKGoldberg:
Fixes #66.

Adds a `/guide/advocacy.html` page under _Resources_ with much of the suggested content from the issue.

### DIFF
--- a/docs/guide/advocacy.md
+++ b/docs/guide/advocacy.md
@@ -1,0 +1,83 @@
+# Advocacy
+
+Much of e18e's work involves making changes in existing open source repositories.
+These repositories generally have their own workflows that their maintainers expect to adhere to.
+It's important when suggesting changes to repositories to respect those workflows and actively work with those maintainers.
+
+## Coordinating Issues
+
+Many repositories prefer discussing potential changes in an issue before they can fully review a pull request.
+Unless a repository maintainer has explicitly indicated a willingness to bypass issue practices, stick with their normal flow of filing an issue and fully filling out the most applicable issue template.
+[Explaining Changes](#explaining-changes) covers the common information you'll want to include.
+
+### Umbrella e18e Tracking
+
+When many changes are possible for a single repository, please search for an "umbrella" issue in [e18e/ecosystem-issues](https://github.com/e18e/ecosystem-issues), and open one if it doesn't already exist.
+That will help with tracking progress and raising visibility for other e18e participants.
+
+We also use these issues to keep track of maintainer preferences around the initiative.
+If a maintainer requests a preference, especially one to not be sent e18e pull requests, please note it there.
+
+## Explaining Changes
+
+Most maintainers are not active participants in e18e.
+Many will have not heard of the initiative in the first place.
+Suggested changes need to include the contextual information that can clearly and quickly explain to a repository maintainer why the change is for the better.
+
+In general, it's best to demonstrate _tangible benefits_ that are _specific to the project_.
+That can change depending on what area(s) of improvement your change falls under.
+
+### Bundle Size
+
+When a project creates a built output such as bundled JS files, and a cleanup is meant to improve the size of those bundles, show the actual file size difference.
+
+This is important because:
+
+* Many changes might change the devDependency tree size, but not impact the bundle
+* Swapping out direct usage of a dependency that is also deeply required might result in a net bigger bundle
+
+In the latter case, you might consider prioritizing working on the upstream consumers of that dependency first.
+
+### Dependency tree
+
+If a dependency cleanup is meant to reduce the size of a project's dependency tree, show the actual change to the tree.
+
+Linking to just the npmgraph.js.org visualization of a package can be informative in isolation but doesn't directly indicate end-user benefits.
+See also [Bundle Size](#bundle-size).
+
+### Speed
+
+If a cleanup is meant to improve the performance of a project, show actual performance measurements for the project.
+Performance-oriented developers generally don't believe any comparisons not backed up with a reproducible experiment.
+
+Isolated benchmarks of a new dependency are not convincing on their own.
+Downstream users often have very different usage scenarios and/or don't use some dependencies in hot paths.
+
+## What Isn't Convincing
+
+Most open source maintainers regularly deal with drive-by contributions that ignore their workflows and do a poor job of explaining themselves.
+Be careful not to make the common mistakes of those contributions, as that weakens the case for your change and for e18e as a whole.
+
+### Abstract Reports
+
+Don't be vague.
+Indicating that a cleanup could offer bug fixes, improved developer experience, more features, or other features in the abstract is not convincing without any evidence those impact the project.
+
+* For bug fixes, be sure to have a reproduction -- much like any other kind of bug report
+* For developer experience and other features, you'll similarly want to be able to demonstrate why the new way is tangibly better
+
+### Downloads and Stars
+
+Non-user-facing metrics such as GitHub Stars and npm download statistics are not generally convincing.
+Both can easily be gamed and thus are not generally seen as reliable metrics.
+
+If your goal is to show that a package is well-known and trusted, it's better to demonstrate the factors that would result in those high numbers.
+Most often that is from showing lists of dependents that include many dependencies and/or high-profile projects.
+
+### Support Ranges
+
+Some cleanups involve restricting project dependency ranges, such as raising a minimum Node.js version from an older version to the current LTS.
+Stricter dependency or platform support ranges on their own are not convincing, user-facing benefits.
+You'll want to instead demonstrate the tangible benefits that come from them.
+
+For example, if you're proposing removing old code, show that the cleanup improves the bundle size, performance, and any other improved metrics.
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -81,6 +81,10 @@ export default defineConfig({
               text: 'Resources',
               link: '/guide/resources',
             },
+            {
+              text: 'Advocacy',
+              link: '/guide/advocacy',
+            },
           ],
         },
         {


docs/guide/advocacy.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Abstract Reports
+
+Don't be vague.
+Indicating that a cleanup could offer bug fixes, improved developer experience, more features, or other features in the abstract is not convincing without any evidence those impact the project.
+
+* For bug fixes, be sure to have a reproduction -- much like any other kind of bug report
+* For developer experience and other features, you'll similarly want to be able to demonstrate why the new way is tangibly better
+
+### Downloads and Stars
```
JoshuaKGoldberg, line 69: Threading https://github.com/e18e/e18e/issues/66#issuecomment-2892488419 (@benmccann):

> I've found myself invoking the npm download stats from time-to-time. Usually when I do so it's because people are worried about stability. My argument is that if a library has a large number of downloads such as 20m downloads / wk then it's got so many users that it's very unlikely there are many undiscovered edge cases. Of course this has to be combined with other information such as how many open issues there are, how quickly they're addressed when discovered, libraries that are using it (to show multiple big consumers and not just one by the same author or faked downloads), etc. Does this make sense or would there be a better way of demonstrating that point?

Yeah, I think when we get to scales like 20m downloads / wk then that is -theoretically- a probable indicator that the package is stable and used. I don't think I've ever seen a package successfully inflate its numbers beyond the thousands [^anecdote].

The problem IME is that so many people have independently tried to promote GitHub stars or npm downloads as a real metric, that many in the community are biased against them. I share this bias: if someone tries to reinforce a package's trustworthiness based on either then I immediately question why they're falling back to that. If the package is so popular, there will certainly be better metrics to show. Why fall back to those marketing-y ones?

tl;dr: IMO if downloads are a good indicator, something else that is more reliable should be too, so we should prefer that something instead

[^anecdote]: I did this recently by accident, haha: https://bsky.app/profile/joshuakgoldberg.com/post/3lmcq5dmlj22q

ljharb: I would say it's important to look at which dependents contribute to that.

If I add package X (with no downloads) to package Y (with 20M downloads), then package X immediately gets a 20M bump, but when the maintainers are different, that's just 1 dependent - the download numbers can be misleading.

and, in the same scenario, when package X is *extracted* from package Y by the same maintainer(s), it _should_ be considered as trustworthy as Y (that trust should ofc not solely be based on download counts)

There's a lot of nuance here.

docs/guide/advocacy.md
JoshuaKGoldberg, line 1: I wanted to include a section in here about social dynamics, but couldn't figure out phrasing. As in, reputation is important for any initiative:

* Clear wins and good behavior paint the initiative in a good light, which helps with convincing folks to go with it
* Going against maintainer practices / doing a poor job of convincing turns into an accidental ["straw man fallacy"](https://en.wikipedia.org/wiki/Straw_man) that turns folks away from the initiative, harming its goals

...maybe this isn't necessary for the guide?

docs/guide/advocacy.md
```diff
@@ -0,0 +1,83 @@
...
+
+## What Isn't Convincing
+
+Most open source maintainers regularly deal with drive-by contributions that ignore their workflows and do a poor job of explaining themselves.
+Be careful not to make the common mistakes of those contributions, as that weakens the case for your change and for e18e as a whole.
+
+### Abstract Reports
+
+Don't be vague.
+Indicating that a cleanup could offer bug fixes, improved developer experience, more features, or other features in the abstract is not convincing without any evidence those impact the project.
```
JoshuaKGoldberg, line 64: 🙈 

```suggestion
Indicating that a cleanup could offer bug fixes, improved developer experience, or other features in the abstract is not convincing without any evidence those impact the project.
```

docs/guide/advocacy.md
```diff
@@ -0,0 +1,83 @@
...
+* Many changes might change the devDependency tree size, but not impact the bundle
+* Swapping out direct usage of a dependency that is also deeply required might result in a net bigger bundle
+
+In the latter case, you might consider prioritizing working on the upstream consumers of that dependency first.
+
+### Dependency tree
+
+If a dependency cleanup is meant to reduce the size of a project's dependency tree, show the actual change to the tree.
+
+Linking to just the npmgraph.js.org visualization of a package can be informative in isolation but doesn't directly indicate end-user benefits.
```
benmccann, line 45: I'm not quite sure what this sentence is suggesting and it sort of sounds to contradict the previous sentence. Is it that PR authors should also explain the degree to which reducing the dependency tree size reduces supply chain risk and saves disk space for users?

docs/guide/advocacy.md
```diff
@@ -0,0 +1,83 @@
...
+See also [Bundle Size](#bundle-size).
+
+### Speed
+
+If a cleanup is meant to improve the performance of a project, show actual performance measurements for the project.
```
benmccann, line 50:
```suggestion
If a cleanup is meant to improve the performance of a project, showing actual performance measurements for the project would be the best option followed by sharing a more general benchmark together with an explanation of why it is relevant such as the benchmarked method being called in a hot code path.
```

docs/guide/advocacy.md
```diff
@@ -0,0 +1,83 @@
...
+When many changes are possible for a single repository, please search for an "umbrella" issue in [e18e/ecosystem-issues](https://github.com/e18e/ecosystem-issues), and open one if it doesn't already exist.
+That will help with tracking progress and raising visibility for other e18e participants.
+
+We also use these issues to keep track of maintainer preferences around the initiative.
+If a maintainer requests a preference, especially one to not be sent e18e pull requests, please note it there.
+
+## Explaining Changes
+
+Most maintainers are not active participants in e18e.
+Many will have not heard of the initiative in the first place.
```
43081j, line 24: this feels like an odd bit of wording to include as it depends on the circle and will change over time

e.g. much of the vue/vite community is aware of e18e, including maintainers. similarly, many of the svelte community's maintainers are too.

it'd probably be better to word this as "Some maintainers are not active participants in e18e or have not heard of it"

docs/guide/advocacy.md
```diff
@@ -0,0 +1,83 @@
...
+In the latter case, you might consider prioritizing working on the upstream consumers of that dependency first.
+
+### Dependency tree
+
+If a dependency cleanup is meant to reduce the size of a project's dependency tree, show the actual change to the tree.
+
+Linking to just the npmgraph.js.org visualization of a package can be informative in isolation but doesn't directly indicate end-user benefits.
+See also [Bundle Size](#bundle-size).
+
+### Speed
```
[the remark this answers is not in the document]
43081j: I know isolated benchmarks don't show the full picture, but they still do show when a package is significantly faster than another.

this is usually the case, so showing the package benchmarks is enough in most of those cases.

it only really matters to do a deep dive when you're comparing two pretty fast packages. then you need to show that it is faster _in this project_ (the target project).

how you do that is up to the maintainer. maybe show some benchmarks of the repo itself, but it will vary.

docs/guide/advocacy.md
```diff
@@ -0,0 +1,83 @@
...
+
+If a cleanup is meant to improve the performance of a project, show actual performance measurements for the project.
+Performance-oriented developers generally don't believe any comparisons not backed up with a reproducible experiment.
+
+Isolated benchmarks of a new dependency are not convincing on their own.
```
43081j, line 53: as per my other comment, this isn't true

in most cases, the replacement package is _significantly_ faster in all cases. so an isolated benchmark _is_ convincing (because its an obvious speed up regardless of which codepath you took).

we shouldn't be telling people isolated benchmarks aren't useful. in the average case, they actually are enough.

when it isn't significantly faster, I agree that a benchmark in the repo should be provided.

JoshuaKGoldberg: > we shouldn't be telling people isolated benchmarks aren't useful

Hmm yeah. I don't want to imply they're not _useful_, just that they need accompanying information. They're definitely useful on their own!

But...

> in the average case, they actually are enough.

I cannot agree with this 😄. Fundamentally, if someone sends me a performance claim, I will _never_ be open to it _unless_ there is an easy reproduction along with it. For really obvious clear-cut cases it might be as simple as _"run the app and see how much faster it is"_. But most of the time I don't believe it's so clear?

I suspect we're actually not that different in opinion here, we're just speaking very differently on the key points around _"convincing"_ and _"isolated benchmarks"_. 

> in most cases, the replacement package is significantly faster.

Asking to learn: can you provide some evidence? I recognize you've been way more entrenched in e18e's movings than me (of course, heh) so if you're seeing this in other packages that's a really interesting learning. My understanding was the vast majority of runtime speed changes are on the scale of shaving a few milliseconds here and there.

JoshuaKGoldberg: > we shouldn't be telling people isolated benchmarks aren't useful

Hmm yeah. I don't want to imply they're not _useful_, just that they need accompanying information. They're definitely useful on their own!

But...

> in the average case, they actually are enough.

I cannot agree with this 😄. Fundamentally, if someone sends me a performance claim, I will _never_ be open to it _unless_ there is an easy reproduction along with it. For really obvious clear-cut cases it might be as simple as _"run the app and see how much faster it is"_. But most of the time I don't believe it's so clear?

I suspect we're actually not that different in opinion here, we're just speaking very differently on the key points around _"convincing"_ and _"isolated benchmarks"_. 

> in most cases, the replacement package is significantly faster.

Asking to learn: can you provide some evidence? I recognize you've been way more entrenched in e18e's movings than me (of course, heh) so if you're seeing this in other packages that's a really interesting learning. My understanding was the vast majority of runtime speed changes are on the scale of shaving a few milliseconds here and there.

43081j: > But most of the time I don't believe it's so clear?

it has been in many, many migrations i've seen in the e18e space. you just so happen to maintain projects that haven't seen this i think

e.g. i've seen projects move to `fdir` and show seconds of speed up in test runs. so its a bit wasteful to spend time doing granular benchmarks and project benchmarks when we can very clearly see the run is significantly faster.

that's why we need to be careful on adding too much burden to PR authors here.

in obvious cases (which i suspect there's a lot of), its unlikely anyone cares for granular benchmarks and in depth project benchmarks. because you can just _run it_ and it'll be obviously faster.

in cases where its not so obvious, benchmark in the repo you're migrating, or just show the difference in run times

a library's own benchmarks are useful on seeing which of these two categories it fits into (is it obviously faster or not). once we have that knowledge, we can decide the next steps based on the maintainers' needs.

but that means we can't discourage any kind of benchmark really since they're all appropriate for different cases

JoshuaKGoldberg: @43081j can I ask you to rewrite this section in the file? I really do think we're pretty close in opinion. +1 from me that if it's straightfoward to demonstrate obvious perf improvements then a benchmark isn't necessary. But I don't know how to phrase this in a way I think you'll agree with. 😄

docs/guide/advocacy.md
43081j:
```suggestion
An isolated dependency graph (e.g. via `npmgraph.js.org`) can be useful to give a quick overview of the difference between two dependencies. However, sometimes it is not representative of how the overall project's dependency tree will be affected. This can be for various reasons, such as the project pulling the same dependencies in elsewhere.

Showing a project dependency graph before and after can be more beneficial for this reason, and may help you find occurrences of target packages you weren't aware of.
```

docs/guide/advocacy.md
43081j:
```suggestion
Isolated benchmarks of a new dependency can be useful when the comparison is large (i.e. you're comparing against a _much_ slower dependency). However, in cases where the gap is small, it is often better to benchmark the project before and after instead of the dependency itself in isolation.

Many maintainers just prefer a project benchmark too, so always refer to the contribution guide or ask the maintainers what their preference is.
```

something like this maybe

JoshuaKGoldberg: This is great, thank you! 🙌

docs/guide/advocacy.md
```diff
@@ -0,0 +1,83 @@
...
+That will help with tracking progress and raising visibility for other e18e participants.
+
+We also use these issues to keep track of maintainer preferences around the initiative.
+If a maintainer requests a preference, especially one to not be sent e18e pull requests, please note it there.
+
+## Explaining Changes
+
+Most maintainers are not active participants in e18e.
+Many will have not heard of the initiative in the first place.
+Suggested changes need to include the contextual information that can clearly and quickly explain to a repository maintainer why the change is for the better.
```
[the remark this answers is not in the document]
43081j: im not sure we need to. happy to just leave it as is, if you are

docs/guide/advocacy.md
```diff
@@ -0,0 +1,83 @@
...
+See also [Bundle Size](#bundle-size).
+
+### Speed
+
+If a cleanup is meant to improve the performance of a project, show actual performance measurements for the project.
+Performance-oriented developers generally don't believe any comparisons not backed up with a reproducible experiment.
+
+Isolated benchmarks of a new dependency are not convincing on their own.
+Downstream users often have very different usage scenarios and/or don't use some dependencies in hot paths.
+
```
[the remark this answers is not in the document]
43081j: it would be nice to end this section in a positive note i agree

much of these sections are "don't do X, Y, or Z". but no positives 😬

docs/guide/advocacy.md
```diff
@@ -0,0 +1,89 @@
+# Advocacy
+
+Much of e18e's work involves making changes in existing open source repositories.
+These repositories generally have their own workflows that their maintainers expect to adhere to.
+It's important when suggesting changes to repositories to respect those workflows and actively work with those maintainers.
+
+## Coordinating Issues
+
+Many repositories prefer discussing potential changes in an issue before they can fully review a pull request.
+Unless a repository maintainer has explicitly indicated a willingness to bypass issue practices, stick with their normal flow of filing an issue and fully filling out the most applicable issue template.
```
benmccann: just a minor wording suggestion as not all maintainers prefer an issue ahead of time
```suggestion
Unless a repository maintainer has explicitly indicated a willingness to bypass issue practices, stick with their normal flow and fully fill out the most applicable issue template if they prefer an issue to be filed ahead of any PRs.
```

docs/guide/advocacy.md
```diff
@@ -0,0 +1,89 @@
...
+Isolated benchmarks of a new dependency can be useful when the comparison is large (i.e. you're comparing against a _much_ slower dependency).
+However, in cases where the gap is small, it is often better to benchmark the project before and after instead of the dependency itself in isolation.
+
+Many maintainers just prefer a project benchmark too, so always refer to the contribution guide or ask the maintainers what their preference is.
+Downstream users often have very different usage scenarios and/or don't use some dependencies in hot paths.
+
+## What Isn't Convincing
+
+Most open source maintainers regularly deal with drive-by contributions that ignore their workflows and do a poor job of explaining themselves.
+Be careful not to make the common mistakes of those contributions, as that weakens the case for your change and for e18e as a whole.
```
benmccann:
```suggestion
Be careful not to make these common mistakes, as that weakens the case for your change and for e18e as a whole.
```

docs/guide/advocacy.md
```diff
@@ -0,0 +1,89 @@
...
+Don't be vague.
+Indicating that a cleanup could offer bug fixes, improved developer experience, or other features in the abstract is not convincing without any evidence those impact the project.
+
+* For bug fixes, be sure to have a reproduction -- much like any other kind of bug report
+* For developer experience and other features, you'll similarly want to be able to demonstrate why the new way is tangibly better
+
+### Downloads and Stars
+
+Non-user-facing metrics such as GitHub Stars and npm download statistics are not generally convincing.
+Both can easily be gamed and thus are not generally seen as reliable metrics.
```
benmccann: "no generally seen as" seems slightly strong. I think they're generally trustworthy and accepted - though I do clearly agree they can be gamed and sometimes are
```suggestion
Both can easily be gamed and thus may not be seen as reliable metrics.
```

docs/guide/advocacy.md
```diff
@@ -0,0 +1,89 @@
...
+
+If your goal is to show that a package is well-known and trusted, it's better to demonstrate the factors that would result in those high numbers.
+Most often that is from showing lists of dependents that include many existing projects, especially other high-profile ones.
+
+### Support Ranges
+
+Some cleanups involve restricting project dependency ranges, such as raising a minimum Node.js version from an older version to the current LTS.
+Stricter dependency or platform support ranges on their own are not convincing, user-facing benefits.
+You'll want to instead demonstrate the tangible benefits that come from them.
```
benmccann: I'd go a bit further than saying that they're not a benefit on their own and say that they're a drawback
```suggestion
Stricter dependency or platform support ranges are a drawback to users as it may cause additional difficulty in upgrading the library.
You'll want to demonstrate that the tangible benefits that come from them are large enough for the majority of users that they they are worth the trade-off.
```

benmccann (commented):
this looks great!  I just took a pass to act as the second pair of eyes James was looking for and only had a few minor comments

docs/guide/advocacy.md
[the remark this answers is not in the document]
outslept: I find mitata to be the most approachable of all the tools, with tinybench being relatively straightforward as well. Both are comparatively simple to get started with, but you still need to understand how to write benchmarks to avoid gaming the results.

- https://github.com/evanwashere/mitata
- https://github.com/tinylibs/tinybench

UPD: Bun docs has an overview on benchmarking: https://bun.sh/docs/project/benchmarking

docs/guide/advocacy.md
```diff
@@ -0,0 +1,89 @@
...
+However, sometimes it is not representative of how the overall project's dependency tree will be affected.
+This can be for various reasons, such as the project pulling the same dependencies in elsewhere.
+
+Showing a project dependency graph before and after can be more beneficial for this reason, and may help you find occurrences of target packages you weren't aware of.
+See also [Bundle Size](#bundle-size).
+
+### Speed
+
+If a cleanup is meant to improve the performance of a project, showing actual performance measurements for the project would be the best option followed by sharing a more general benchmark together with an explanation of why it is relevant such as the benchmarked method being called in a hot code path.
+Performance-oriented developers generally don't believe any comparisons not backed up with a reproducible experiment.
```
outslept: don't believe sounds dismissive

```suggestion
Performance-oriented developers generally expect comparisons to include reproducible experiments.
```

docs/guide/advocacy.md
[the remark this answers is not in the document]
JoshuaKGoldberg: > you still need to understand how to write benchmarks to avoid gaming the results

Yeah, good point. I'm worried about developers using low-level libraries in those exact accidentally-gaming ways. IME low-level benchmarks are pretty easy to mess up like that.

I added a few lines in the latest commit, 8e1215a. How's that?

43081j:
@JoshuaKGoldberg if you're happy with it, im happy to merge and share this

just let me know

looks like all other significant comments have been resolved

JoshuaKGoldberg:
Sudden zoom in effect on a Squirrel yelling. Caption: "SHIP IT!"

PR #77, opened 2025-08-22
docs: add replacement guides

43081j:


docs/guide/replacement-guides/deep-equal.md
```diff
@@ -0,0 +1,38 @@
...
+Example:
+
+```ts
+import equal from 'deep-equal' // [!code --]
+import dequal from 'dequal' // [!code ++]
+
+const a = { foo: 'bar' }
+const b = { foo: 'bar' }
+
+equal(a, b) // true // [!code --]
```
outslept: Looks like a typo.

now: 

Image

after:

Image

```suggestion
equal(a, b) // true [!code --]
```

docs/guide/replacement-guides/deep-equal.md
```diff
@@ -0,0 +1,38 @@
...
+dequal(a, b) // true // [!code ++]
```
outslept: same

```suggestion
dequal(a, b) // true [!code ++]
```

docs/guide/replacement-guides/deep-equal.md
```diff
@@ -0,0 +1,38 @@
...
+
+```ts
+import fastDeepEqual from 'fast-deep-equal' // [!code ++]
+import deepEqual from 'deep-equal' // [!code --]
+
+const a = { foo: 'bar' }
+const b = { foo: 'bar' }
+
+deepEqual(a, b) // true // [!code --]
+fastDeepEqual(a, b) // true // [!code ++]
```
outslept:
```suggestion
deepEqual(a, b) // true [!code --]
fastDeepEqual(a, b) // true [!code ++]
```

outslept:
little nit: maybe there's an approach with sidebar? 

e.g. vueuse's; for them it is

image

for e18e we could do something like:

image

because currently we lose touch with sidebar and can't navigate between pages easily

43081j:
i agree, it'd be nice to present this a bit less hidden-in-the-depths. but we also don't want a giant sidebar of all modules

ill merge as is for now and we can figure it out later 👍

PR #78, opened 2025-08-25
docs: add more packages to replacements guide

outslept:
this pr adds

- `is-builtin-module`
- `emoji-regex`
- `eslint-plugin-vitest`
- `md5`
- `eslint-plugin-node`

to the replacement guide

### DIFF
--- a/docs/guide/replacement-guides/emoji-regex.md
+++ b/docs/guide/replacement-guides/emoji-regex.md
@@ -13,12 +13,12 @@ const text = `
 \u{2194}\u{FE0F}: ↔️ default text presentation character rendered as emoji
 \u{1F469}: 👩 emoji modifier base (Emoji_Modifier_Base)
 \u{1F469}\u{1F3FF}: 👩🏿 emoji modifier base followed by a modifier
-`;
+`
 
-const regex = emojiRegex();
+const regex = emojiRegex()
 for (const match of text.matchAll(regex)) {
-  const emoji = match[0];
-  console.log(`Matched sequence ${emoji} — code points: ${[...emoji].length}`);
+  const emoji = match[0]
+  console.log(`Matched sequence ${emoji} — code points: ${[...emoji].length}`)
 }
 ```
 
@@ -29,11 +29,11 @@ If your target runtime supports ES2024 Unicode property sets, you can use the na
 ```ts
 import emojiRegex from 'emoji-regex' // [!code --]
 
-const regex = emojiRegex(); // [!code --]
+const regex = emojiRegex() // [!code --]
 const regex = /\p{RGI_Emoji}/gv // [!code ++]
 
 for (const match of text.matchAll(regex)) {
-  const emoji = match[0];
-  console.log(`Matched sequence ${emoji} — code points: ${[...emoji].length}`);
+  const emoji = match[0]
+  console.log(`Matched sequence ${emoji} — code points: ${[...emoji].length}`)
 }
 ```
--- a/docs/guide/replacement-guides/eslint-plugin-node.md
+++ b/docs/guide/replacement-guides/eslint-plugin-node.md
@@ -5,7 +5,7 @@
 [`eslint-plugin-n`](https://github.com/eslint-community/eslint-plugin-n) is a direct fork which is actively maintained. It has new features, bugfixes and updated dependencies.
 
 ```js
-import nPlugin from 'eslint-plugin-n'       // [!code ++]
+import nPlugin from 'eslint-plugin-n' // [!code ++]
 import nodePlugin from 'eslint-plugin-node' // [!code --]
 
 export default [
@@ -13,13 +13,13 @@ export default [
     files: ['**/*.js'], // or any other pattern
     plugins: {
       node: nodePlugin, // [!code --]
-      n: nPlugin,       // [!code ++]
+      n: nPlugin, // [!code ++]
     },
     rules: {
       ...nodePlugin.configs['recommended-script'].rules, // [!code --]
-      ...nPlugin.configs['recommended-script'].rules,    // [!code ++]
+      ...nPlugin.configs['recommended-script'].rules, // [!code ++]
       'node/exports-style': ['error', 'module.exports'], // [!code --]
-      'n/exports-style': ['error', 'module.exports'],    // [!code ++]
+      'n/exports-style': ['error', 'module.exports'], // [!code ++]
     },
   },
 ]
@@ -30,9 +30,9 @@ If you're using a legacy config format:
 ```js
 module.exports = {
   extends: [
-    "eslint:recommended",
-    "plugin:node/recommended", // [!code --]
-    "plugin:n/recommended",    // [!code ++]
+    'eslint:recommended',
+    'plugin:node/recommended', // [!code --]
+    'plugin:n/recommended', // [!code ++]
   ],
-};
+}
 ```
--- a/docs/guide/replacement-guides/is-builtin-module.md
+++ b/docs/guide/replacement-guides/is-builtin-module.md
@@ -41,4 +41,3 @@ import { builtinModules } from 'node:module'
 
 console.log(builtinModules)
 ```
-


docs/guide/replacement-guides/emoji-regex.md
```diff
@@ -0,0 +1,43 @@
...
+```ts
+import emojiRegex from 'emoji-regex' // [!code --]
+
+const regex = /\p{RGI_Emoji}/gv // [!code ++]
+
+for (const match of text.matchAll(regex)) {
+  const emoji = match[0];
+  console.log(`Matched sequence ${emoji} — code points: ${[...emoji].length}`);
+}
+```
```
outslept: I was unsure if we need this. Mentioned in the docs for [emoji-regex-xs](https://github.com/slevithan/emoji-regex-xs#comparison-with-emoji-regex-and-prgi_emoji) and has it's own quirks

docs/guide/replacement-guides/emoji-regex.md
```diff
@@ -0,0 +1,43 @@
...
+// original regular expression.
+
+const text = `
+\u{231A}: ⌚ default emoji presentation character (Emoji_Presentation)
+\u{2194}\u{FE0F}: ↔️ default text presentation character rendered as emoji
+\u{1F469}: 👩 emoji modifier base (Emoji_Modifier_Base)
+\u{1F469}\u{1F3FF}: 👩🏿 emoji modifier base followed by a modifier
+`;
+
+const regex = emojiRegex();
```
43081j: do we need a diff here too?

```ts
const regex = emojiRegex; // [!code --]
const regex = emojiRegex(); // [!code ++]
```

docs/guide/replacement-guides/emoji-regex.md
43081j: i think its right to mention `RGI_Emoji` 👍

docs/guide/replacement-guides/emoji-regex.md
outslept: no? that assigns the function reference, not a RegExp obj; don't think this is what we want

docs/guide/replacement-guides/eslint-plugin-node.md
```diff
@@ -0,0 +1,35 @@
...
+  },
+]
+```
+
+If you're using a legacy JSON config:
+
+```json
+{
+  "extends": ["eslint:recommended", "plugin:node/recommended"] // [!code --]
+  "extends": ["eslint:recommended", "plugin:n/recommended"]    // [!code ++]
```
43081j: the diff may look nicer if we do this:

```json
{
  "extends": [
    "eslint:recommended",
    "plugin:node/recommended", // [!code --]
    "plugin:n/recommended", // [!code ++]
  }
}
```

and maybe it should be `json5` to allow for comments?

docs/guide/replacement-guides/eslint-plugin-vitest.md
```diff
@@ -0,0 +1,22 @@
...
+    files: ['tests/**'], // or any other pattern
+    plugins: {
+      vitest,
+    },
+    rules: {
+      ...vitest.configs.recommended.rules, // you can also use vitest.configs.all.rules to enable all rules
+      'vitest/max-nested-describe': ['error', { max: 3 }], // you can also modify rules' behavior using option like this
+    },
+  },
+]
```
43081j, line 22: missing the closing backticks here

docs/guide/replacement-guides/eslint-plugin-node.md
outslept: I'll rewrite this to js format I guess as eslint is being noisy otherwise

docs/guide/replacement-guides/is-builtin-module.md
```diff
@@ -0,0 +1,22 @@
+# Replacements for `is-builtin-module`
+
+## Node.js (since 16.x)
```
43081j, line 3: we actually had `builtinModules` from way back in Node 6 iirc (but not `isBuiltinModule`)

so maybe we should say `since 6.x` but have two examples: one for before 16.x and one for after

not sure how best to organise that though

docs/guide/replacement-guides/emoji-regex.md
43081j: i understood that, but i misunderstood that _both_ packages export a function

this note confused me:

```ts
// Note: because the regular expression has the global flag set, this module
// exports a function that returns the regex rather than exporting the regular
// expression itself, to make it impossible to (accidentally) mutate the
// original regular expression.
```

i took this as a note you wrote to tell us the new one exports a function, and the old one didn't. but both do, so maybe we don't need this note?

docs/guide/replacement-guides/is-builtin-module.md
outslept: oh really, I never knew that. from what I understand for older version you need to implement your own check something like

```js
import { builtinModules } from 'module'

function isBuiltin(moduleName) {
  const name = moduleName.startsWith('node:') 
    ? moduleName.slice(5) 
    : moduleName
  
  return builtinModules.includes(name)
}

isBuiltin('fs') // true
```

outslept: I'll check how Sindre handled this historically

43081j: exactly what you wrote is how to do it

`is-builtin-module` only really had a use pre-6.x (if it even existed then)

we could say `since 6.x` but leave a sub-heading like:

### Before 16.x

Before Node 16.x, `isBuiltinModule` was not available, so you would have to use the modules list directly:

```ts
// ..
```

or maybe better, have a separate heading `Node.js (since 6.x)`

outslept: sure, I'll file a PR in es-tooling/module-replacements as well

docs/guide/replacement-guides/is-builtin-module.md
```diff
@@ -0,0 +1,44 @@
...
+
+function isBuiltin(moduleName) { // [!code ++]
+  const name = moduleName.startsWith('node:') // [!code ++]
+    ? moduleName.slice(5) // [!code ++]
+    : moduleName // [!code ++]
+
+  return builtinModules.includes(name) // [!code ++]
+} // [!code ++]
+
+isBuiltin('fs') // true // [!code ++]
```
43081j: you can remove the 2nd `//` (so it becomes `// true [!code ++]`
and the same below

docs/guide/replacement-guides/is-builtin-module.md
```diff
@@ -0,0 +1,44 @@
...
+isBuiltinModule('fs') // true // [!code --]
+```
+
+## Getting all built-in modules
+
+For a full list of built-in modules (available since Node.js 6.x), you can use [builtinModules](https://nodejs.org/api/module.html#modulebuiltinmodules):
+
+```ts
+import { builtinModules } from 'node:module' // [!code ++]
```
43081j:
```suggestion
import { builtinModules } from 'node:module'
```

docs/guide/replacement-guides/emoji-regex.md
```diff
@@ -0,0 +1,44 @@
...
+for (const match of text.matchAll(regex)) {
+  const emoji = match[0];
+  console.log(`Matched sequence ${emoji} — code points: ${[...emoji].length}`);
+}
+```
+
+## `\p{RGI_Emoji}`
```
43081j: lets actually name this one `Unicode RegExp (native)` i think

outslept: Maybe Unicode RegExp (`\p{RGI_Emoji}`) sound better?

docs/guide/replacement-guides/is-builtin-module.md
outslept: oh I remember suggesting removing this in your pr :D

docs/guide/replacement-guides/emoji-regex.md
outslept: Changed to Unicode Property Escape

docs/guide/replacement-guides/emoji-regex.md
outslept: yep, I've copied the example from emoji-regex's docs and the note comes from there. I'll remove it

docs/guide/replacement-guides/emoji-regex.md
43081j: it'd be nice to follow the pattern we have in other pages of adding `(native)`

it is regexp specific, which is why i was trying to mention that

basically `RGI_Emoji` is an implementation detail, really the solution is the fact we're using a regexp (e.g. you could use the `Emoji` property too to match things like keycap emojis)

43081j: maybe something like `RegExp unicode properties (native)`?

outslept: I agree. I'll go for with suggestion

PR #80, opened 2025-08-30
docs: add more packages to replacements guide

outslept:
this PR adds:

- `eslint-plugin-es`
- `eslint-plugin-eslint-comments`
- `npm-run-all`
- `rimraf`
- `strip-ansi`

to the replacements guide

### DIFF
--- a/docs/guide/replacement-guides/eslint-plugin-es.md
+++ b/docs/guide/replacement-guides/eslint-plugin-es.md
@@ -2,7 +2,7 @@
 
 ## `eslint-plugin-es-x`
 
-eslint-plugin-es-x is a direct fork which is actively maintained. It has new features, bugfixes and updated dependencies.
+[eslint-plugin-es-x](https://github.com/eslint-community/eslint-plugin-es-x) is a direct fork which is actively maintained. It has new features, bugfixes and updated dependencies.
 
 ```js
 import pluginESx from "eslint-plugin-es-x" // [!code ++]
@@ -26,7 +26,7 @@ module.exports = {
     'es-x'   // [!code ++]
   ],
   rules: {
-    // 'es/no-regexp-lookbehind-assertions': 'error', // [!code --]
+    'es/no-regexp-lookbehind-assertions': 'error', // [!code --]
     'es-x/no-regexp-lookbehind-assertions': 'error', // [!code ++]
   }
 }
--- a/docs/guide/replacement-guides/eslint-plugin-es.md
+++ b/docs/guide/replacement-guides/eslint-plugin-es.md
@@ -1,0 +1,33 @@
+# Replacements for `eslint-plugin-es`
+
+## `eslint-plugin-es-x`
+
+eslint-plugin-es-x is a direct fork which is actively maintained. It has new features, bugfixes and updated dependencies.
+
+```js
+import pluginESx from "eslint-plugin-es-x" // [!code ++]
+
+export default [
+    pluginESx.configs['flat/restrict-to-es2018'],
+]
+```
+
+If you're using a legacy config format:
+
+```js
+module.exports = {
+  extends: [
+    'eslint:recommended',
+    'plugin:es/restrict-to-es2018', // [!code --]
+    'plugin:es-x/restrict-to-es2018', // [!code ++]
+  ],
+  plugins: [
+    'es',    // [!code --]
+    'es-x'   // [!code ++]
+  ],
+  rules: {
+    // 'es/no-regexp-lookbehind-assertions': 'error', // [!code --]
+    'es-x/no-regexp-lookbehind-assertions': 'error', // [!code ++]
+  }
+}
+```
--- a/docs/guide/replacement-guides/eslint-plugin-es.md
+++ b/docs/guide/replacement-guides/eslint-plugin-es.md
@@ -5,10 +5,30 @@
 [eslint-plugin-es-x](https://github.com/eslint-community/eslint-plugin-es-x) is a direct fork which is actively maintained. It has new features, bugfixes and updated dependencies.
 
 ```js
+import { FlatCompat } from '@eslint/eslintrc' // [!code --]
+import pluginES from 'eslint-plugin-es' // [!code --]
 import pluginESx from 'eslint-plugin-es-x' // [!code ++]
 
+const compat = new FlatCompat() // [!code --]
+
 export default [
-  pluginESx.configs['flat/restrict-to-es2018'], // [!code ++]
+  {
+    files: ['**/*.js'],
+    languageOptions: {
+      ecmaVersion: 2020,
+    },
+    plugins: {
+      'es': pluginES, // [!code --]
+      'es-x': pluginESx, // [!code ++]
+    },
+    rules: {
+      'es/no-regexp-lookbehind-assertions': 'error', // [!code --]
+      'es-x/no-regexp-lookbehind-assertions': 'error' // [!code ++]
+    },
+  },
+
+  ...compat.extends('plugin:es/restrict-to-es2018'), // [!code --]
+  pluginESx.configs['flat/restrict-to-es2018'],
 ]
 ```
 
--- a/docs/guide/replacement-guides/eslint-plugin-es.md
+++ b/docs/guide/replacement-guides/eslint-plugin-es.md
@@ -8,7 +8,7 @@ eslint-plugin-es-x is a direct fork which is actively maintained. It has new fea
 import pluginESx from "eslint-plugin-es-x" // [!code ++]
 
 export default [
-    pluginESx.configs['flat/restrict-to-es2018'],
+    pluginESx.configs['flat/restrict-to-es2018'], // [!code ++]
 ]
 ```
 
--- a/docs/guide/replacement-guides/eslint-plugin-es.md
+++ b/docs/guide/replacement-guides/eslint-plugin-es.md
@@ -28,7 +28,7 @@ export default [
   },
 
   ...compat.extends('plugin:es/restrict-to-es2018'), // [!code --]
-  pluginESx.configs['flat/restrict-to-es2018'],
+  pluginESx.configs['flat/restrict-to-es2018'], // [!code ++]
 ]
 ```
 
--- a/docs/guide/replacement-guides/eslint-plugin-es.md
+++ b/docs/guide/replacement-guides/eslint-plugin-es.md
@@ -5,10 +5,10 @@
 [eslint-plugin-es-x](https://github.com/eslint-community/eslint-plugin-es-x) is a direct fork which is actively maintained. It has new features, bugfixes and updated dependencies.
 
 ```js
-import pluginESx from "eslint-plugin-es-x" // [!code ++]
+import pluginESx from 'eslint-plugin-es-x' // [!code ++]
 
 export default [
-    pluginESx.configs['flat/restrict-to-es2018'], // [!code ++]
+  pluginESx.configs['flat/restrict-to-es2018'], // [!code ++]
 ]
 ```
 
@@ -22,8 +22,8 @@ module.exports = {
     'plugin:es-x/restrict-to-es2018', // [!code ++]
   ],
   plugins: [
-    'es',    // [!code --]
-    'es-x'   // [!code ++]
+    'es', // [!code --]
+    'es-x' // [!code ++]
   ],
   rules: {
     'es/no-regexp-lookbehind-assertions': 'error', // [!code --]
--- a/docs/guide/replacement-guides/eslint-plugin-eslint-comments.md
+++ b/docs/guide/replacement-guides/eslint-plugin-eslint-comments.md
@@ -5,17 +5,20 @@
 [`@eslint-community/eslint-plugin-eslint-comments`](https://github.com/eslint-community/eslint-plugin-eslint-comments) is the actively maintained successor with updated dependencies, flat config support, and continued development.
 
 ```js
-import comments from '@eslint-community/eslint-plugin-eslint-comments/configs' // [!code ++]
-import js from '@eslint/js'
+import commentsCommunity from '@eslint-community/eslint-plugin-eslint-comments/configs' // [!code ++]
+import eslintComments from 'eslint-plugin-eslint-comments' // [!code --]
 
 export default [
-  js.configs.recommended,
-  comments.recommended, // [!code ++]
+  commentsCommunity.recommended, // [!code ++]
   {
+    plugins: {
+      'eslint-comments': eslintComments, // [!code --]
+    },
     rules: {
+      'eslint-comments/no-unused-disable': 'error',         // [!code --]
       '@eslint-community/eslint-comments/no-unused-disable': 'error', // [!code ++]
     }
-  },
+  }
 ]
 ```
 
--- a/docs/guide/replacement-guides/eslint-plugin-eslint-comments.md
+++ b/docs/guide/replacement-guides/eslint-plugin-eslint-comments.md
@@ -5,8 +5,9 @@
 [`@eslint-community/eslint-plugin-eslint-comments`](https://github.com/eslint-community/eslint-plugin-eslint-comments) is the actively maintained successor with updated dependencies, flat config support, and continued development.
 
 ```js
-import commentsCommunity from '@eslint-community/eslint-plugin-eslint-comments/configs' // [!code ++]
 import eslintComments from 'eslint-plugin-eslint-comments' // [!code --]
+// eslint-disable-next-line perfectionist/sort-imports
+import commentsCommunity from '@eslint-community/eslint-plugin-eslint-comments/configs' // [!code ++]
 
 export default [
   commentsCommunity.recommended, // [!code ++]
--- a/docs/guide/replacement-guides/eslint-plugin-eslint-comments.md
+++ b/docs/guide/replacement-guides/eslint-plugin-eslint-comments.md
@@ -15,7 +15,7 @@ export default [
       'eslint-comments': eslintComments, // [!code --]
     },
     rules: {
-      'eslint-comments/no-unused-disable': 'error',         // [!code --]
+      'eslint-comments/no-unused-disable': 'error', // [!code --]
       '@eslint-community/eslint-comments/no-unused-disable': 'error', // [!code ++]
     }
   }
@@ -26,14 +26,14 @@ If you're using a legacy config format:
 
 ```js
 module.exports = {
-    extends: [
-        "eslint:recommended",
-        "plugin:eslint-comments/recommended", // [!code --]
-        "plugin:@eslint-community/eslint-comments/recommended" // [!code ++]
-    ],
-    rules: {
-        "eslint-comments/no-unused-disable": "error", // [!code --]
-        "@eslint-community/eslint-comments/no-unused-disable": "error" // [!code ++]
-    }
+  extends: [
+    'eslint:recommended',
+    'plugin:eslint-comments/recommended', // [!code --]
+    'plugin:@eslint-community/eslint-comments/recommended' // [!code ++]
+  ],
+  rules: {
+    'eslint-comments/no-unused-disable': 'error', // [!code --]
+    '@eslint-community/eslint-comments/no-unused-disable': 'error' // [!code ++]
+  }
 }
 ```
--- a/docs/guide/replacement-guides/eslint-plugin-eslint-comments.md
+++ b/docs/guide/replacement-guides/eslint-plugin-eslint-comments.md
@@ -1,0 +1,36 @@
+Replacements for `eslint-plugin-eslint-comments`
+
+## `@eslint-community/eslint-plugin-eslint-comments`
+
+[`@eslint-community/eslint-plugin-eslint-comments`](https://github.com/eslint-community/eslint-plugin-eslint-comments) is the actively maintained successor with updated dependencies, flat config support, and continued development.
+
+```js
+import js from "@eslint/js"
+import comments from "@eslint-community/eslint-plugin-eslint-comments/configs" // [!code ++]
+
+export default [
+    js.configs.recommended,
+    comments.recommended, // [!code ++]
+    {
+        rules: {
+            "@eslint-community/eslint-comments/no-unused-disable": "error", // [!code ++]
+        }
+    },
+]
+```
+
+If you're using a legacy config format:
+
+```js
+module.exports = {
+    extends: [
+        "eslint:recommended",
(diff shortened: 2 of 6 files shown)


docs/guide/replacement-guides/strip-ansi.md
```diff
@@ -0,0 +1,37 @@
...
+console.log(stripVTControlCharacters('\u001B[4me18e\u001B[0m')); // [!code ++]
+```
+
+## Bun
+
+[Bun supports the same API](https://bun.sh/reference/node/util/stripVTControlCharacters#node:util.stripVTControlCharacters)
+
+```js
+import { stripVTControlCharacters } from 'node:util' // [!code ++]
+import stripAnsi from 'strip-ansi' // [!code --]
```
Jayllyz, line 46: Bun can use this now
https://bun.com/reference/bun/stripANSI

outslept: Thank you, I didn't know that ❤️ Are there any notable differences between stripAnsi and stripVTControlCharacters in their implementation?

outslept: Aha! I've found https://github.com/oven-sh/bun/pull/21801. Really cool that Jared and team care about things like this

(+ https://bun.com/blog/release-notes/bun-v1.2.21#bun-stripansi-simd-accelerated-ansi-escape-removal)

docs/guide/replacement-guides/strip-ansi.md
```diff
@@ -0,0 +1,45 @@
+# Replacements for `strip-ansi`
+
+## Node.js
+
+Added in v16.11.0, [util.stripVTControlCharacters](https://nodejs.org/api/util.html#utilstripvtcontrolcharactersstr) can be used to strip ANSI escape codes from a string.
```
outslept, line 5: Perhaps we should mention this as well? https://github.com/nodejs/node/issues/53697

docs/guide/replacement-guides/eslint-plugin-es.md
```diff
@@ -0,0 +1,33 @@
+# Replacements for `eslint-plugin-es`
+
+## `eslint-plugin-es-x`
+
+[eslint-plugin-es-x](https://github.com/eslint-community/eslint-plugin-es-x) is a direct fork which is actively maintained. It has new features, bugfixes and updated dependencies.
+
+```js
+import pluginESx from 'eslint-plugin-es-x' // [!code ++]
```
43081j, line 10: we need to show the code from before, too

for example

```ts
import pluginES from 'eslint-plugin-es'; // [!code --]
```

and the same for the config

outslept: If I remember correctly the code is not identical for Flat Config there and I didn't want to create a separate fenced code block

docs/guide/replacement-guides/eslint-plugin-eslint-comments.md
```diff
@@ -0,0 +1,36 @@
+# Replacements for `eslint-plugin-eslint-comments`
+
+## `@eslint-community/eslint-plugin-eslint-comments`
+
+[`@eslint-community/eslint-plugin-eslint-comments`](https://github.com/eslint-community/eslint-plugin-eslint-comments) is the actively maintained successor with updated dependencies, flat config support, and continued development.
+
+```js
+import comments from '@eslint-community/eslint-plugin-eslint-comments/configs' // [!code ++]
```
43081j: this too should show a before/after, and in the following code

docs/guide/replacement-guides/npm-run-all.md
```diff
@@ -0,0 +1,57 @@
+# Replacements for `npm-run-all`
+
+## `npm-run-all2`
+
+[npm-run-all2](https://github.com/bcomnes/npm-run-all2) is an actively maintained fork with important fixes, dependency updates. The CLI is fully compatible, so you can migrate by replacing the package name.
```
43081j: > so you can migrate by replacing the package.

otherwise it sounds like you need to replace the _name_ in your scripts

docs/guide/replacement-guides/npm-run-all.md
```diff
@@ -0,0 +1,57 @@
...
+The commands remain the same: `npm-run-all`, `run-s`, and `run-p`.
+
+## `concurrently`
+
+Another option is [concurrently](https://github.com/open-cli-tools/concurrently), which focuses on running scripts in parallel with colored output and process control. It uses a slightly different syntax but works well for replacing the `--parallel` use case.
+
+```json
+{
+  "scripts": {
+    "dev": "concurrently \"npm:watch-*\" \"npm:start\""
```
43081j: this needs a diff too

docs/guide/replacement-guides/rimraf.md
```diff
@@ -0,0 +1,43 @@
+# Replacements for `rimraf`
+
+## Node.js
+
+If you are using [`rimraf`](https://github.com/isaacs/rimraf) to remove files and directories, modern Node.js provides native alternatives. Since Node v14.14.0, [`fs.rm`](https://nodejs.org/api/fs.html#fspromisesrmpath-options) supports recursive deletion and works as a direct replacement.
```
43081j: we don't need this first bit since anyone reaching this doc is definitely using rimraf. it can be simplified to something like:

> Node.js v14.14.0 and above provide a native alternative [...]

docs/guide/replacement-guides/rimraf.md
```diff
@@ -0,0 +1,43 @@
...
+
+```js
+import { rm } from 'node:fs/promises' // [!code ++]
+import rimraf from 'rimraf' // [!code --]
+
+await rimraf('./dist') // [!code --]
+await rm('./dist', { recursive: true, force: true }) // [!code ++]
+```
+
+## Node.js (older versions)
```
43081j: instead of `older versions`, it might be clearer to be explicit - which versions? `Node.js (before x.y.z)` for example

docs/guide/replacement-guides/rimraf.md
```diff
@@ -0,0 +1,43 @@
...
+```
+
+---
+
+## `premove`
+
+For cases where you cannot rely on Node being available, use [`premove`](https://github.com/lukeed/premove). It includes a CLI and works on Node v8 and newer.
+
+```sh
+premove ./dist
```
43081j: this one should show a diff too

docs/guide/replacement-guides/rimraf.md
```diff
@@ -0,0 +1,43 @@
...
+
+```sh
+node -e "require('fs').rmSync('./dist', { recursive: true, force: true, maxRetries: process.platform === 'win32' ? 10 : 0 })"
+```
+
+---
+
+## `premove`
+
+For cases where you cannot rely on Node being available, use [`premove`](https://github.com/lukeed/premove). It includes a CLI and works on Node v8 and newer.
```
43081j: the use case is more that you're on an older node or you need a CLI, rather than node itself not being available

docs/guide/replacement-guides/eslint-plugin-es.md
outslept: Yep.

```js
// eslint.config.mjs

// @ts-check

import pluginES from 'eslint-plugin-es'
import pluginESx from 'eslint-plugin-es-x'

export default [
  {
    files: ['**/*.js'],
    languageOptions: {
      ecmaVersion: 2020,
    },
    plugins: {
      es: pluginES,
      'es-x': pluginESx,
    },
    rules: {
      'es/no-regexp-lookbehind-assertions': 'error',
      'es-x/no-regexp-named-capture-groups': 'error',
    },
  },

  pluginES.configs['restrict-to-es2018'],
  pluginESx.configs['flat/restrict-to-es2019'],
]
```

If you run `eslint --inspect-config` with this config you'll see `ConfigError: Config (unnamed): Unexpected key "extends" found.` Do we need to mention that this plugin does not support eslint 9 (for the same reason we recommend switching to an alternative)?

docs/guide/replacement-guides/strip-ansi.md
```diff
@@ -0,0 +1,45 @@
...
+console.log(stripAnsi('\u001B[4me18e\u001B[0m')) // [!code --]
+console.log(stripVTControlCharacters('\u001B[4me18e\u001B[0m')) // [!code ++]
+```
+
+## Bun
+
+Bun provides two options:
+
+1. [`util.stripVTControlCharacters`](https://bun.sh/reference/node/util/stripVTControlCharacters) is supported through Bun’s Node API layer.
+2. [Since Bun v1.2.21, `Bun.stripANSI`](https://bun.com/blog/release-notes/bun-v1.2.21#bun-stripansi-simd-accelerated-ansi-escape-removal) offers a high‑performance, SIMD‑accelerated alternative, often 6×–57× faster than strip-ansi.
```
43081j: i don't think we need to mention the "faster than" since every alternative is probably 6x faster than `strip-ansi`, and we don't mention it in the others

i'd just say its a faster alternative to `util`

docs/guide/replacement-guides/strip-ansi.md
```diff
@@ -0,0 +1,45 @@
...
+
+```js
+// Node-compatible API
+import { stripVTControlCharacters } from 'node:util' // [!code ++]
```
43081j: the diff in this code block doesn't make much sense:
- the comments look as if they were always there
- two replacements are being shown at once

i think you need to split this up into two sub-sections of the bun section, and each has its own code block with a proper diff. the bullets above go away then and just go in their own section

docs/guide/replacement-guides/eslint-plugin-eslint-comments.md
outslept: Luckily this is not the same as the case above. I'll handle it

outslept: Resolved 👍🏼

docs/guide/replacement-guides/npm-run-all.md
outslept: 👍🏼

docs/guide/replacement-guides/npm-run-all.md
outslept: 👍🏼

docs/guide/replacement-guides/rimraf.md
outslept: 👍🏼

docs/guide/replacement-guides/rimraf.md
outslept: 👍🏼

docs/guide/replacement-guides/rimraf.md
outslept: 👍🏼

docs/guide/replacement-guides/rimraf.md
outslept: not really like the way i phrase it, but 👍🏼 (open to suggestions)

docs/guide/replacement-guides/strip-ansi.md
outslept: 👍🏼

docs/guide/replacement-guides/strip-ansi.md
outslept: 👍🏼

docs/guide/replacement-guides/strip-ansi.md
43081j: they're so rarely used, i don't think we need to. if they ever become more of a thing, we can update the docs

docs/guide/replacement-guides/eslint-plugin-es.md
43081j: are you saying that eslint-plugin-es doesn't support flat configs?

if that's the case, you should remove the diff

this is a diff as soon as you add `[!code ++]`. what are you diffing against if there's no previous code?

docs/guide/replacement-guides/eslint-plugin-eslint-comments.md
```diff
@@ -0,0 +1,39 @@
+# Replacements for `eslint-plugin-eslint-comments`
+
+## `@eslint-community/eslint-plugin-eslint-comments`
+
+[`@eslint-community/eslint-plugin-eslint-comments`](https://github.com/eslint-community/eslint-plugin-eslint-comments) is the actively maintained successor with updated dependencies, flat config support, and continued development.
+
+```js
+import commentsCommunity from '@eslint-community/eslint-plugin-eslint-comments/configs' // [!code ++]
+import eslintComments from 'eslint-plugin-eslint-comments' // [!code --]
```
43081j, line 8: the removed code should come before the added code (like it does further down)

docs/guide/replacement-guides/npm-run-all.md
```diff
@@ -0,0 +1,62 @@
...
+
+## `npm-run-all2`
+
+[npm-run-all2](https://github.com/bcomnes/npm-run-all2) is an actively maintained fork with important fixes, dependency updates.
+
+```json
+{
+  "devDependencies": {
+    "npm-run-all": "^X.Y.Z", // [!code --]
+    "npm-run-all2": "^X.Y.Z" // [!code ++]
```
43081j: we don't need to show the dependency diff, just the usage

this one doesn't need a diff since its the same before and after (but does need the code example ofc)

docs/guide/replacement-guides/npm-run-all.md
```diff
@@ -0,0 +1,62 @@
...
+
+## `concurrently`
+
+Another option is [concurrently](https://github.com/open-cli-tools/concurrently), which focuses on running scripts in parallel with colored output and process control. It uses a slightly different syntax but works well for replacing the `--parallel` use case.
+
+```json
+{
+  "scripts": {
+    "dev:npm-run-all": "npm-run-all --parallel \"watch-*\" start", // [!code --]
+    "dev:concurrently": "concurrently \"npm:watch-*\" \"npm:start\"" // [!code ++]
```
43081j: we should name this the same script. in real world it wouldn't be called `dev:npm-run-all` or `dev:concurrently`. 

remember this is a migration example, so it'd be the same name. you wouldn't rename the script as part of the migration

e.g. call it `dev:watch` or something, or just `dev`. whatever you'd call it in a real project you'd actually setup

docs/guide/replacement-guides/rimraf.md
```diff
@@ -0,0 +1,44 @@
+# Replacements for `rimraf`
+
+## Node.js
+
+Node.js v14.14.0 and above provide a native alternative: [`fs.rm`](https://nodejs.org/api/fs.html#fspromisesrmpath-options). It supports recursive deletion and works as a direct replacement.
+
+```js
+import { rm } from 'node:fs/promises' // [!code ++]
+import rimraf from 'rimraf' // [!code --]
```
43081j, line 8: the removal should come first like the other diffs

docs/guide/replacement-guides/rimraf.md
```diff
@@ -0,0 +1,44 @@
...
+await rm('./dist', { recursive: true, force: true }) // [!code ++]
+```
+
+## Node.js (before v14.14.0)
+
+If you need to support Node.js 12 up to 14.13, you can use [`fs.rmdir`](https://nodejs.org/api/fs.html#fsrmdirpath-options-callback) with the recursive option. This was added in Node v12.10.0, though it’s deprecated as of Node v14.
+
+```js
+import { rmdir } from 'node:fs/promises'
+import rimraf from 'rimraf' // [!code --]
```
43081j, line 20: here too, the removal should come first
and `rmdir` should be an addition

docs/guide/replacement-guides/rimraf.md
```diff
@@ -0,0 +1,44 @@
...
+
+```sh
+node -e "require('fs').rmSync('./dist', { recursive: true, force: true, maxRetries: process.platform === 'win32' ? 10 : 0 })"
+```
+
+---
+
+## `premove`
+
+If you are on an older Node.js version (before v12.10) or you specifically need a CLI replacement, use [`premove`](https://github.com/lukeed/premove). It provides both an API and a CLI and works on Node.js v8 and newer.
```
43081j: > If you are on an older Node.js version (before v12.10) or you specifically need a CLI replacement, you can use [...]

instead of saying `use premove`, `you can use premove` (since this whole doc is a list of suggested replacements rather than us telling you to use one or the other)

docs/guide/replacement-guides/rimraf.md
```diff
@@ -0,0 +1,44 @@
...
+
+```sh
+rimraf ./dist # [!code --]
+premove ./dist # [!code ++]
```
43081j: we should show this in an npm script i think, e.g.:

```json
{
  "scripts": {
    "clean": "premove lib"
  }
}
```

43081j:
overall comment: we need to make sure we're consistently ordering diffs

in some we have `add/remove`, in others `remove/add`, etc

outslept:
> overall comment: we need to make sure we're consistently ordering diffs
> 
> in some we have `add/remove`, in others `remove/add`, etc

hmm.. some ordering is done via eslint but it doesn't account for our chosen orden. lemme see what we can do with this

docs/guide/replacement-guides/npm-run-all.md
outslept: I was fighting eslint this way. I'll disable the rule 👍🏼

docs/guide/replacement-guides/eslint-plugin-es.md
outslept: I'll rewrite using @eslint/compat package

docs/guide/replacement-guides/rimraf.md
outslept: 👍🏼

docs/guide/replacement-guides/npm-run-all.md
outslept: 👍🏼

docs/guide/replacement-guides/rimraf.md
outslept: 👍🏼

docs/guide/replacement-guides/rimraf.md
outslept: 👍🏼

docs/guide/replacement-guides/eslint-plugin-eslint-comments.md
outslept: 👍🏼

docs/guide/replacement-guides/npm-run-all.md
outslept: 👍🏼

docs/guide/replacement-guides/eslint-plugin-es.md
outslept: 👍🏼

docs/guide/replacement-guides/npm-run-all.md
```diff
@@ -0,0 +1,58 @@
...
+}
+```
+
+The commands remain the same: `npm-run-all`, `run-s`, and `run-p`.
+
+## `concurrently`
+
+Another option is [concurrently](https://github.com/open-cli-tools/concurrently), which focuses on running scripts in parallel with colored output and process control. It uses a slightly different syntax but works well for replacing the `--parallel` use case.
+
+```json
```
43081j, line 21: this should probably be `json5` or `ts` now (so it allows comments). not sure if json5 exists in vitepress

docs/guide/replacement-guides/rimraf.md
```diff
@@ -0,0 +1,48 @@
...
+await rm('./dist', { recursive: true, force: true }) // [!code ++]
+```
+
+## Node.js (before v14.14.0)
+
+If you need to support Node.js 12 up to 14.13, you can use [`fs.rmdir`](https://nodejs.org/api/fs.html#fsrmdirpath-options-callback) with the recursive option. This was added in Node v12.10.0, though it’s deprecated as of Node v14.
+
+```js
+import rimraf from 'rimraf' // [!code --]
+import { rmdir } from 'node:fs/promises'
```
43081j: this is an addition

docs/guide/replacement-guides/rimraf.md
```diff
@@ -0,0 +1,48 @@
...
+
+## CLI usage
+
+To replace `rimraf` inside npm scripts, you can run Node directly in eval mode:
+
+```sh
+node -e "require('fs').rmSync('./dist', { recursive: true, force: true, maxRetries: process.platform === 'win32' ? 10 : 0 })"
+```
+
+---
```
43081j: we don't need this separator

docs/guide/replacement-guides/strip-ansi.md
```diff
@@ -0,0 +1,51 @@
+# Replacements for `strip-ansi`
```
43081j, line 1: this one just needs the add/remove ordering like the others but seems good other than that

outslept: When I was trying this the last time, jsonc or json5 did nothing 😅 (I rewrote legacy eslint config format in replacement docs from json to js because of it in the other PR).

docs/guide/replacement-guides/npm-run-all.md
outslept: When I was trying this the last time, jsonc or json5 did nothing 😅 (I rewrote legacy eslint config format in replacement docs from json to js because of it in the other PR).

outslept: Wait. It does not raise an error for me? (I think the plugin does some sort of magic under the hood and counts JSON with comments as JSONc automatically. Like we use tsconfig.json (it has comments most of the time), but the extension is still .json)

image

docs/guide/replacement-guides/strip-ansi.md
outslept: 👍🏼

docs/guide/replacement-guides/rimraf.md
outslept: 👍🏼

docs/guide/replacement-guides/rimraf.md
outslept: 👍🏼

43081j (approved):
looks good to me!

docs/guide/replacement-guides/npm-run-all.md
43081j: i think we can leave it 👍

PR #87, opened 2025-09-03
docs(chalk): add note about colors in browsers

jycouet:
Feel free to tweak the message 👍

### DIFF
--- a/docs/guide/replacement-guides/chalk.md
+++ b/docs/guide/replacement-guides/chalk.md
@@ -93,7 +93,6 @@ console.log(
 ```
 
 Library support:
-- [`chalk`](https://github.com/chalk/chalk#browser-support) - colors are supported in _Chromium_ browsers
 - [`ansis`](https://github.com/webdiscus/ansis#browser-compatibility-for-ansi-codes) - colors are supported in _Chromium_ browsers
 - `picocolors` - strips colors in browser environments
 - `node:util` - is not available in browsers


outslept:
I’m a bit hesitant to recommend ansis outright for Chromium DevTools without context. This page **is terminal‑focused now**, while browser consoles use a different mechanism (`%c` + CSS via `console.log`). Chromium DevTools do render ANSI SGR sequences, but that behavior is not standardized and is not supported across browsers - there's an ongoing discussion in WHATWG/Console repository about it.

If we want to keep a browser note, I’d probably move it to the end under a dedicated heading like "Browser support" (so the terminal/Node guidance stays front and center) and there give a short, focused breakdown of browser behavior and links.

docs/guide/replacement-guides/chalk.md
```diff
@@ -74,3 +74,26 @@ Similarly, you can use RGB and hex colors:
...
+  'color: #ec8f5e;',
+  'color: #f2ca60;',
+  'color: #bece57;',
+  'color: #7bb560;',
+  'Ecosystem Performance'
+)
+```
+
+Library support:
+- `chalk` - colors are supported in _Chromium_ browsers
```
outslept:
```suggestion
- [`chalk`](https://github.com/chalk/chalk#browser-support) - colors are supported in _Chromium_ browsers
```

docs/guide/replacement-guides/chalk.md
```diff
@@ -74,3 +74,26 @@ Similarly, you can use RGB and hex colors:
...
+- `ansis` - colors are supported in _Chromium_ browsers
```
outslept:
```suggestion
- [`ansis`](https://github.com/webdiscus/ansis#browser-compatibility-for-ansi-codes) - colors are supported in _Chromium_ browsers
```

outslept:
Can't target it through a suggestion, but we probably have to adapt meta description as well

Maybe something like

```
description: Modern alternatives to the chalk package for terminal string styling and colors, with notes on browser console support
```

Is good?

jycouet:
The best part 😜
image

Should be good.

docs/guide/replacement-guides/chalk.md
```diff
@@ -74,3 +74,26 @@ Similarly, you can use RGB and hex colors:
...
+  'Hello %ce%c1%c8%ce',
+  'color: #ec8f5e;',
+  'color: #f2ca60;',
+  'color: #bece57;',
+  'color: #7bb560;',
+  'Ecosystem Performance'
+)
+```
+
+Library support:
```
43081j, line 95: i'm not sure we need this list, especially listing chalk since this doc is reached by users looking to _drop_ chalk

maybe we just need to mention something like:

> If you need a package which works both in terminals and in browsers, consider using ansis.

since it seems no other package supports browsers?

jycouet: Maybe dropping `chalk` of the list is sufficient?

I think that it's important to note somewhere that `node:util` will not work in browser. It's maybe obvious for a lot of people. But as a user looking at replacements, I see `chalk`, cool, I can follow e18e suggestion. So I take the first item `node:util`, do the switch... and my app is not working anymore in some pages.
So, to me, it's worth mentioning it.

43081j: that's fair 👍  just dropping chalk works

43081j (approved):

PR #91, opened 2025-09-08
feat: add `shortid`

outslept:


docs/guide/replacement-guides/shortid.md
```diff
@@ -0,0 +1,44 @@
...
+
+```ts
+shortid.characters('0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ$@') // [!code --]
+import { customAlphabet } from 'nanoid' // [!code ++]
+
+const alphabet = '0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ$@' // [!code ++]
+const makeId = customAlphabet(alphabet, 12) // [!code ++]
+const id = makeId() // [!code ++]
+```
+
```
outslept, line 44: I was unsure if we need to include Web Crypto API here (https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID). What do you think?

43081j: i think we can leave it like this for now 👍  if we change our mind later we can add to the doc

PR #97, opened 2025-09-08
docs: add `sort-object`

outslept:
I didn't know how deep we go with providing code samples so tried to keep it minimal.

### DIFF
--- a/docs/guide/replacement-guides/sort-object.md
+++ b/docs/guide/replacement-guides/sort-object.md
@@ -1,0 +1,63 @@
+---
+description: Modern alternatives to the sort-object package for sorting object keys
+---
+
+# Replacements for `sort-object`
+
+## JavaScript APIs (`Object.keys` + `Array.sort`)
+
+For simple cases:
+
+```ts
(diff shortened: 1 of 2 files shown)


docs/guide/replacement-guides/sort-object.md
```diff
@@ -0,0 +1,63 @@
...
+## JavaScript APIs (`Object.keys` + `Array.sort`)
+
+For simple cases:
+
+```ts
+import sortObj from 'sort-object' // [!code --]
+const sorted = sortObj(object) // [!code --]
+
+// Ascending A→Z
+const sorted = Object.keys(object) // [!code ++]
```
43081j: i wonder if we should use `Object.fromEntries` instead

```ts
Object.fromEntries(
  Object.entries(obj).sort((a, b) => a[0].localeCompare(b[0]))
);
```

it might satisfy typescript more than keying into the object

outslept: 👍🏼

PR #98, opened 2025-09-09
docs: add `fast-glob` replacements

outslept:


docs/guide/replacement-guides/fast-glob.md
```diff
@@ -0,0 +1,29 @@
...
+const files = await fg('**/*.ts', { // [!code --]
+  cwd: process.cwd(), // [!code --]
+  ignore: ['**/node_modules/**'] // [!code --]
+}) // [!code --]
+
+const files = await glob('**/*.ts', { // [!code ++]
+  cwd: process.cwd(), // [!code ++]
+  ignore: ['**/node_modules/**'], // [!code ++]
+  expandDirectories: false // [!code ++]
+}) // [!code ++]
```
43081j:
```suggestion
const files = await fg('**/*.ts', { // [!code --]
const files = await glob('**/*.ts', { // [!code ++]
  cwd: process.cwd(),
  ignore: ['**/node_modules/**'],
  expandDirectories: false // [!code ++]
})
```

linter won't be happy since its invalid JS but its a nicer diff

outslept: I didn't like this option (updated the file according to it tho), because this doesn't feel right

image

(breaks syntax highlighting for the addition)

We have the same "problem" here https://e18e.dev/guide/replacement-guides/glob.html#fs-glob-native-since-node-22-x

43081j: it breaks syntax highlighting but thats because linters and syntax highlighters think its a regular code block

its actually a diff and this syntactically makes sense for a diff

i suppose the vitepress integration should force us to use a custom language like `tsdiff` instead of `ts`

PR #101, opened 2025-09-09
docs: add `mkdirp` replacements

outslept:


docs/guide/replacement-guides/mkdirp.md
```diff
@@ -0,0 +1,59 @@
...
+
+# Replacements for `mkdirp` / `make-dir`
+
+# Alternatives
+
+## Node.js (since v10.12.0)
+
+Node.js v10.12.0 and up supports the `recursive` option in the [`fs.mkdir`](https://nodejs.org/api/fs.html#fsmkdirpath-options-callback) function, which allows parent directories to be created automatically:
+
+```js
```
43081j: Maybe this first example isn't necessary? Since we show the diff afterwards which is far more useful I think

outslept:
> I probably missed Bun API for this. I'll check tomorrow

Briefly checked now. Doesn't seem to be the case. Bun has `Bun.write`, but it allows writing files, not simply creating dirs

43081j:
for multi-module docs like this (mkdirp, make-dir) - i wonder if we should just document one of them even if we send people to this doc for similar modules

if the examples were different enough, we'd have separate docs i think

the fact they can share the same doc tells me the diff will be almost identical

docs/guide/replacement-guides/mkdirp.md
```diff
@@ -0,0 +1,41 @@
...
+
+From [`make-dir`](https://github.com/sindresorhus/make-dir):
+
+```js
+import { makeDirectory, makeDirectorySync } from 'make-dir' // [!code --]
+import { mkdir, mkdirSync } from 'node:fs' // [!code ++]
+import { mkdir as mkdirAsync } from 'node:fs/promises' // [!code ++]
+
+// Async
+await makeDirectory('unicorn/rainbow/cake') // [!code --]
```
43081j: lets get rid of the unicorn remnants here too

outslept: 👍🏼

43081j (approved):

PR #105, opened 2025-09-10
docs: add `uri-js` replacements

outslept:


docs/guide/replacement-guides/uri-js.md
```diff
@@ -0,0 +1,62 @@
...
+- Node.js URL: https://nodejs.org/api/url.html#class-url
+
+Example:
+
+```ts
+import * as URI from 'uri-js' // [!code --]
+URI.resolve('https://a/b/c/d?q', '../../g') // [!code --]
+
+/* no import required for URL */
+new URL('../../g', 'https://a/b/c/d?q').href // [!code ++]
```
43081j:
```suggestion
import * as URI from 'uri-js' // [!code --]

URI.resolve('https://a/b/c/d?q', '../../g') // [!code --]
new URL('../../g', 'https://a/b/c/d?q').href // [!code ++]
```

docs/guide/replacement-guides/uri-js.md
```diff
@@ -0,0 +1,62 @@
...
+## `fast-uri`
+
+[`fast-uri`](https://github.com/fastify/fast-uri) is a zero-dependency, high-performance RFC 3986 URI toolbox (parse/serialize/resolve/equal) with options similar to `uri-js`.
+
+```ts
+import * as URI from 'uri-js' // [!code --]
+import * as uri from 'fast-uri' // [!code ++]
+
+URI.parse('uri://user:pass@example.com:123/one/two.three?q1=a1#a') // [!code --]
+uri.parse('uri://user:pass@example.com:123/one/two.three?q1=a1#a')  // [!code ++]
```
43081j: if we just call both `uri`, the diff will be reduced, i think?

in that only the import has to change

docs/guide/replacement-guides/uri-js.md
outslept: 👍🏼

docs/guide/replacement-guides/uri-js.md
outslept: 👍🏼

PR #109, opened 2025-09-12
fix: use default process import instead of named one

outslept:
image

Without this you'd get

### DIFF
--- a/scripts/add-replacement.js
+++ b/scripts/add-replacement.js
@@ -1,5 +1,6 @@
 import { existsSync, readFileSync, writeFileSync } from 'node:fs'
 import { dirname, join } from 'node:path'
+import process from 'node:process'
 import { createInterface } from 'node:readline'
 import { fileURLToPath } from 'node:url'
 


scripts/add-replacement.js
```diff
@@ -1,6 +1,5 @@
 import { existsSync, readFileSync, writeFileSync } from 'node:fs'
 import { dirname, join } from 'node:path'
-import { process } from 'node:process'
```
43081j, line 3: it does actually exist and is a cleaner way than relying on globals

i think the problem is that its a default export, not a named one (should be `import process`)

outslept: default import works, but why do we need it here?

43081j: its a preference thing but also cleaner than relying on globals

e.g. a linter needs to be configured to have `process` as a global

its just nice to import it like everything else and have a proper dependency graph rather than relying on a global

outslept: 👍🏼

PR #113, opened 2025-09-12
docs: add `dot-prop` replacements

outslept:
currently suffers from the same problem as in here:

https://github.com/e18e/e18e/pull/101#issuecomment-3273746734

### DIFF
--- a/docs/docs/replacements/dot-prop.md
+++ b/docs/docs/replacements/dot-prop.md
@@ -20,22 +20,6 @@ setProperty(obj, 'foo.bar.baz', 'value') // [!code --]
 dset(obj, 'foo.bar.baz', 'value') // [!code ++]
 ```
 
-## `get-value` + `set-value`
-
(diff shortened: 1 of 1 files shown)


docs/guide/replacement-guides/dot-prop.md
```diff
@@ -0,0 +1,58 @@
...
+import { dset } from 'dset' // [!code ++]
+
+const value = getProperty(obj, 'foo.bar.baz') // [!code --]
+const value = delve(obj, 'foo.bar.baz') // [!code ++]
+
+setProperty(obj, 'foo.bar.baz', 'value') // [!code --]
+dset(obj, 'foo.bar.baz', 'value') // [!code ++]
+```
+
+## `get-value` + `set-value`
```
43081j: I'm not sure we should be recommending these

They do exactly the same job as dlv/dset which are arguably better. 

I think we should only be recommending multiple alternatives if they offer something different

outslept: Removed 👍🏼

PR #115, opened 2025-09-13
docs: add `body-parser` replacements

outslept:


docs/guide/replacement-guides/body-parser.md
```diff
@@ -6,4 +6,41 @@ description: Modern alternatives to the body-parser package for parsing HTTP req
...
+
+app.use(json()) // [!code ++]
+app.use(urlencoded()) // [!code ++]
+```
+
+For API differences and feature comparison, see the [migration.md](https://github.com/tinyhttp/milliparsec/blob/master/migration.md).
+
+## Built-in Express parsers
+
+Express 4.16.0+ includes built-in body parsing middleware based on [`body-parser`](https://github.com/expressjs/body-parser).
```
outslept: Maybe I should mention that this uses [expressjs/body-parser module](https://www.npmjs.com/package/body-parser) module underneath?

Ref: https://expressjs.com/en/changelog/ (for 4.16.0)

43081j: yeah this is a weird one because express still uses body-parser under the hood

we're directing people here to drop body-parser, not just hide it away

im not sure we should have this section at all until the day express allows you to bring your own parser

outslept: 👍🏼

PR #116, opened 2025-09-13
docs: add `depcheck` replacements 

outslept:


docs/guide/replacement-guides/depcheck.md
```diff
@@ -0,0 +1,18 @@
...
+# Replacements for `depcheck`
+
+## `knip`
+
+[knip](https://github.com/webpro-nl/knip) is a more actively maintained and feature-rich alternative to [`depcheck`](https://github.com/depcheck/depcheck).
+
+Example:
+
+```json
+{
```
43081j, line 14: maybe since these configs wont really cross over, we should just show a basic knip config?

or point out that it usually doesn't need config

outslept: Initially I made 2 diffs:

1 with scripts (e.g.):

```
npx depcheck
npx knip
```

2 was this diff I left

Want me to stick with the first one and I can show a basic config afterwards I guess?

43081j: what i mean is not have a diff at all, since they're not equivalent in any way. they're two different tools to do the same job, rather than knip being the successor

so we should probably just show a basic example knip config or explain that you don't usually need one

outslept: 👍🏼

PR #120, opened 2025-09-18
chore: update dependencies and lint

bluwy:
The lint was also failing in `main`. This PR fixes them altogether.

Honestly, I wonder if we should remove the eslint setup altogether and replace with just prettier, or just nothing, if it wasn't relied upon so far.


docs/package.json
```diff
@@ -8,8 +8,8 @@
     "serve": "vitepress serve"
   },
   "devDependencies": {
-    "@shikijs/vitepress-twoslash": "^2.3.2",
-    "feed": "^4.2.2",
-    "vitepress": "^1.6.3"
+    "@shikijs/vitepress-twoslash": "^3.12.2",
+    "feed": "^5.1.0",
+    "vitepress": "^2.0.0-alpha.12"
```
bluwy, line 13: This uses the alpha version like from the vite repo. The alpha version uses the latest vite version and has been stable enough so far.

43081j: seems fine to me 👍

PR #124, opened 2025-09-18
docs: add `tempy` replacements

outslept:
Deno APIs were mentioned as early as v1.x (current version: 2.5.1), but I couldn't find their exact introduction date. (e.g. `Deno.makeTempFile` was mentioned here - https://deno.com/blog/v1)

Node example and article link came from here - https://github.com/es-tooling/module-replacements/blob/main/docs/modules/tempy.md


docs/guide/replacement-guides/tempy.md
```diff
@@ -0,0 +1,77 @@
...
+const tempDir = temporaryDirectory() // [!code --]
+
+import {mkdtemp, realpath} from 'node:fs/promises' // [!code ++]
+import {join} from 'node:path' // [!code ++]
+import {tmpdir} from 'node:os' // [!code ++]
+
+// macOS and some other platforms may return a symlink from `os.tmpdir`.
+// For some applications, this can cause problems; thus, we use `realpath`.
+const tempDir = await mkdtemp(join(await realpath(tmpdir()), 'foo-')) // [!code ++]
+```
```
43081j:
````suggestion
```js
import {temporaryDirectory} from 'tempy' // [!code --]
import {mkdtemp, realpath} from 'node:fs/promises' // [!code ++]
import {join} from 'node:path' // [!code ++]
import {tmpdir} from 'node:os' // [!code ++]

const tempDir = temporaryDirectory() // [!code --]
const tempDir = await mkdtemp(join(await realpath(tmpdir()), 'foo-')) // [!code ++]
```
````
we can probably also drop the note about realpath or move it into an actual `!NOTE` markdown

File: docs/guide/replacement-guides/tempy.md [hunk 2]
```diff
@@ -0,0 +1,77 @@
...
+
+Directory cleanup can be done by passing `{recursive: true}` to [`fs.rm`](https://nodejs.org/api/fs.html#fsrmpath-options-callback), available in v14.14.0 and up:
+
+```js
```
43081j: this section doesn't need to be a diff since we're not going from anything to anything

its just some info (that we can use `recursive: true`)

docs/guide/replacement-guides/tempy.md
outslept: removed it 👍🏼

docs/guide/replacement-guides/tempy.md
outslept: 👍🏼

docs/docs/replacements/tempy.md
```diff
@@ -0,0 +1,80 @@
...
+}) // [!code ++]
+
+try {
+  await writeFile(join(tempDir, 'bar.txt'), 'Hello, world!', { encoding: 'utf-8' })
+  await writeFile(join(tempDir, 'baz.txt'), '...', { encoding: 'utf-8' })
+}
+ finally {
+  await rm(tempDir, { recursive: true })
+}
+```
```
43081j:
````suggestion
```js
import { rm, writeFile } from 'node:fs/promises'
import { join } from 'node:path'

// logic to create tempDir...

try {
  await writeFile(join(tempDir, 'bar.txt'), 'Hello, world!', { encoding: 'utf-8' })
  await writeFile(join(tempDir, 'baz.txt'), '...', { encoding: 'utf-8' })
}
 finally {
  await rm(tempDir, { recursive: true })
}
```
````

the diff was a bit broken, so we may as well just drop it i think

**Comment by outslept** [comment]
ah wait a minute :D

File: docs/docs/replacements/tempy.md [hunk 4]
```diff
@@ -0,0 +1,59 @@
...
+
+```ts
+import { temporaryDirectory } from 'tempy'
+ // [!code --]
+const tempDir = temporaryDirectory({ prefix: 'foo-' }) // [!code --]
+
+const tempDir = await Deno.makeTempDir({ prefix: 'foo-' }) // [!code ++]
+
+try { // [!code ++]
+  await Deno.writeTextFile(`${tempDir}/bar.txt`, 'Hello, world!') // [!code ++]
```
43081j: this diff doesn't make sense since there wasn't any code before other than the `tempDir` creation

would the try/finally have existed before _and_ after? maybe just need to remove the diffs from this bit?

outslept: Yeah, I'll remove try..catch

outslept: Should be resolved

PR #128, opened 2025-09-19
docs: move some sections around

43081j:
Moves to this structure:

- Learn
   - Learn about e18e and the general idea
- Docs
  - All docs of our tools, guides, etc
- Blog

@bluwy can you help me figure out how to redirect the old ones? and double check my changes here


43081j:
you say that like we've made a mistake 😂 

this is why the PR is here! i'll update 😄

bluwy:
The redirects can be setup [like this](https://github.com/vitejs/vite/blob/3ef03ab2c75066911bfe40c9b7b1119e53f9bcf3/docs/public/_redirects#L19). I think something like this as a start should work:

```
/guide/replacement-guides/* /docs/replacement-guides/* 301
/guide/* /learn/* 301
```

Also, I wonder if `/docs/replacements` should be `/docs/replacements/` and `/docs/replacement-guides/*` be `/docs/replacements/*`? Might be cleaner and similar to the learn pages.

bluwy:
> i cant get the redirects working locally 🤔 any idea? i pushed what you said

I think you can only verify that once published to netlify since vitepress doesn't have the netlify-specific syntax support.

Looks like you can test it with the deploy preview site and I might've typoed the setup. Should be:

```
/guide/replacement-guides/* /docs/replacement-guides/:splat 301
/guide/* /learn/:splat 301
```

PR #129, opened 2025-09-19
docs: add `execa` replacements

outslept:
https://github.com/es-tooling/module-replacements/blob/main/docs/modules/process-exec.md

### DIFF
--- a/docs/docs/replacements/globby.md
+++ b/docs/docs/replacements/globby.md
@@ -1,0 +1,42 @@
+---
+description: Modern alternatives to the globby package for globbing and .gitignore support
+---
+
+# Replacements for `globby`
+
+## `tinyglobby`
+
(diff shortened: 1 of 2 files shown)


docs/guide/replacement-guides/execa.md
```diff
@@ -0,0 +1,50 @@
...
+
+```ts
+import { execa } from 'execa' // [!code --]
+import exec from 'nanoexec' // [!code ++]
+
+const { stdout } = await execa('echo', ['example']) // [!code --]
+const res = await exec('echo', ['example']) // [!code ++]
+const stdout = res.stdout.toString('utf8') // [!code ++]
+
+await exec('echo "example"', { shell: true }) // [!code ++]
```
43081j: this one isn't in the previous code from what i can see

outslept: ahh, let me fix this

outslept: 👍🏼

PR #133, opened 2025-09-19
docs: add publishing guide

43081j:
This is the branch we can use for building up the publishing guide.

Those of you collaborating on this, please make branches/PRs which target this one or PR suggestions I can accept in.

People working on this one: @jviide @sxzz @dominikg

### DIFF
--- a/docs/docs/publishing.md
+++ b/docs/docs/publishing.md
@@ -44,7 +44,7 @@ This ensures that nobody can push directly to the `main` branch, and that all ch
 
 Navigate to `Settings > Secrets & Variables > Actions`:
 
-- Remove any existing npm secrets from repository secrets.
+- Remove all unnecessary repository secrets (e.g. npm tokens, as OIDC can be used instead)
 
 ### Setting up Trusted Publishing
 
@@ -58,6 +58,9 @@ In this "Trusted Publishing" section, setup a trusted publisher for the workflow
 - Workflow filename is the name of the workflow file you'll create (e.g. `publish.yml`)
 
 While you're in there, also check the box for "Require two-factor authentication and disallow tokens (recommended)". This will ensure that manual publishing _must_ use 2FA.
+
+> [!TIP]
+> It can be a slow job opening all of your packages individually and changing these settings. To assist with this, you can use the [open-packages-on-npm](https://github.com/antfu/open-packages-on-npm) tool in your local repository to open the package(s) on npm, each in a new tab. You can then use this [userscript](https://github.com/sxzz/userscripts/blob/main/src/npm-trusted-publisher.md) to quickly update the trusted publisher settings on each page.
 
 ## Standard Workflow
 
--- a/docs/docs/publishing.md
+++ b/docs/docs/publishing.md
@@ -10,9 +10,20 @@ In this document, we will cover a typical workflow with security in mind, and so
 
 ## Prerequisites
 
+### Enforcing 2FA
+
+Before anything, it is important to enable 2FA in your various accounts and use a secure authentication method.
+
+In your npm account, navigate to the `Account` page and ensure you have 2FA enabled. Ideally, use a security key, otherwise use an authenticator app.
+
+Similarly, in your GitHub account, navigate to `Settings > Password and authentication` and ensure you have 2FA enabled. Again, ideally use a security key, otherwise use an authenticator app.
+
+> [!NOTE]
+> A security key isn't necessarily a physical key (e.g. YubiKey), it can also be something your OS provides, like Windows Hello or macOS Touch ID.
+
 ### Setting up the GitHub repository
 
-First thing is to ensure your GitHub repository is setup securely.
+Next, we need to ensure your GitHub repository is setup securely.
 
 #### Actions Settings
 
--- a/docs/docs/publishing.md
+++ b/docs/docs/publishing.md
@@ -1,0 +1,61 @@
+# Publishing Packages
+
+As a maintainer, there are many options available to you on how you setup your publishing workflow. This may be manual, semi-automated, or fully automated.
+
+In this document, we will cover a typical workflow with security in mind, and some recipes for more advanced workflows.
+
+## Standard Workflow
+
+TODO: Explain the standard workflow. This should roughly be:
+
+- One workflow (GitHub action)
+- Pin actions to a hash
+- Trigger on tag or GitHub release publish
+- Build the package and `npm publish` in the same job
+  - Possibly disable scripts even in CI (but this would mess up things like esbuild, so maybe not)
+- Setup OIDC in npm
+- Disable tokens in npm
+- Require 2FA in npm and GitHub for you and your organisation where possible
+- Use a GitHub environment with required reviewers for the publish job
+- Setup dependabot or renovate to keep dependencies up to date
+
+### Note on local publishing
+
+TODO: Make sure we mention in a sub-section that it may still be worth doing local 2FA publishes if you're a solo maintainer _until_ GitHub supports 2FA for workflow approvals.
+
+Basically, if you're a solo maintainer it may be safer to publish locally with 2FA for now.
+
+If you're part of an org, or just want to use workflows, certainly do not publish locally.
+
+## Changesets Workflow
+
+TODO: Explain roughly the same as the "Standard workflow", but using `changesets` to manage versions.
+
+- Show a diagram of the flow possibly
+- Use the same steps as the standard flow mostly
+
+## Tips & Tools
+
+TODO: Show some further tips and tools from the community.
+
+Briefly explain each of these:
+
+- [`actions-up`](https://github.com/azat-io/actions-up) for managing updates of GitHub actions
+- [`zizmor`](https://github.com/zizmorcore/zizmor) for linting GitHub workflows
+- [`publint`](https://github.com/publint/publint) for linting that your publish config is right
+- [`taze`](https://github.com/antfu-collective/taze) for managing dependency updates manually
+- [`multiocular`](https://github.com/multiocular-com/multiocular) for visualising diffs of dependency changes
+
+Either a table or a bullet list will probably work as long as we have a description, name, and link.
+
+## Further Security
+
+TODO: Explain further security measures that can be taken. Each as a sub-section
+
+Probably these:
+
+- Using physical security keys for 2FA
+  - Or webauthn if that isn't possible
+- Remove all npm tokens from your account
+- Add protection rules to branches and tags
+- Enable immutable releases
--- a/docs/docs/publishing.md
+++ b/docs/docs/publishing.md
@@ -6,34 +6,91 @@ In this document, we will cover a typical workflow with security in mind, and so
 
 ## Standard Workflow
 
-TODO: Explain the standard workflow. This should roughly be:
+With the rise of supply-chain attacks, it is more important than ever to ensure that your publishing workflow is secure. The following is the typical workflow we recommend for most maintainers.
 
-- One workflow (GitHub action)
-- Pin actions to a hash
-- Trigger on tag or GitHub release publish
-- Build the package and `npm publish` in the same job
-  - Possibly disable scripts even in CI (but this would mess up things like esbuild, so maybe not)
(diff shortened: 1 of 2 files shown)


docs/docs/publishing.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Standard Workflow
+
+TODO: Explain the standard workflow. This should roughly be:
+
+- One workflow (GitHub action)
+- Pin actions to a hash
+- Trigger on tag or GitHub release publish
+- Build the package and `npm publish` in the same job
+  - Possibly disable scripts even in CI (but this would mess up things like esbuild, so maybe not)
```
43081j: lets leave a `NOTE` section when we write this

we should recommend disabling scripts in CI and locally. but the note should explain that some dependencies may rely on postinstall scripts, especially those which pull or build native bindings

docs/docs/publishing.md
```diff
@@ -0,0 +1,211 @@
...
+
+#### Actions Settings
+
+Navigate to `Settings > Actions > General` and configure:
+
+- ✓ Require actions to be pinned to a full-length commit SHA
+- ✓ Require approval for first-time contributors
+- ✓ Set default workflow permissions to "Read repository contents and packages permissions"
+
+These settings ensure that outside contributors cannot run arbitrary code in your workflows without your approval first, and that all actions used in your workflows are pinned to a specific commit SHA, preventing supply-chain attacks through action updates.
```
jviide, line 38: Maybe worth a note that the settings can also be set on an organization level too for easier time with multi-repo projects?

docs/docs/publishing.md
```diff
@@ -0,0 +1,211 @@
...
+
+#### Branch Protection
+
+Navigate to `Settings > Branches` and configure:
+
+- ✓ Create a ruleset for your default branch (e.g. `main`)
+- ✓ Require a pull request before merging
```
jviide, line 49: Maybe we should recommend enabling some or all of the following, maybe even others:
  * Require approval
  * Dismiss stale pull request approvals when new commits are pushed
  * Require approval of the most recent reviewable push

Otherwise an attacker might be able to smuggle a nasty commit after the PR has been reviewed but not yet merged. For example see https://www.wiz.io/blog/github-actions-security-guide#branch-protection-28.

docs/docs/publishing.md
```diff
@@ -0,0 +1,211 @@
...
+
+- Test (runs tests)
+- Build (builds the package)
+- Publish (publishes the already built package)
+
+There are a few important things to note about this workflow:
+
+- The package is built in a separate job to publishing, ensuring the publish permissions are not exposed to runtime build code
+- All actions are pinned to a full-length commit SHA
+- Install is done with `--ignore-scripts` to avoid running any lifecycle scripts that may be malicious
```
jviide, line 96: Though the focus is on the publishing phase, it might be beneficial if the `ignore-scripts=true` setting is set in the .npmrc file to block install scripts from running on the developer's machine.

Maybe too much of a tidbit, but could be worth pointing out as I still see lots of `npm install` calls instead of `npm ci` in workflows: `npm ci` requires the package lock to be in sync with package.json, otherwise it barfs. `npm install` on the other hand just dutifully "updates" package lock if needed, potentially installing new and compromised versions of packages.

docs/docs/publishing.md
```diff
@@ -0,0 +1,211 @@
...
+
+- You must still push tags manually
+- Any time a tag is pushed, a GitHub release is created with a changelog generated from commit messages
+- Any time a tag is pushed, the package is published
+
+## Maintenance & Tooling
+
+### Managing dependency updates
+
+We also highly recommend setting up [dependabot](https://docs.github.com/en/code-security/dependabot) or [renovate](https://www.whitesourcesoftware.com/free-developer-tools/renovate/) to keep your dependencies up to date.
```
jviide, line 156: Renovate and dependabot also support keeping actions up to date, so people already using them won't necessary need `actions-up`.

docs/docs/publishing.md
```diff
@@ -0,0 +1,211 @@
...
+
+You can use [`multiocular`](https://github.com/multiocular-com/multiocular) to visualise changes in your dependencies between versions.
+
+This will give you a visualisation of what code has changed when a dependency updated, which can help you identify potential security issues and breaking changes.
+
+## Further Security
+
+### Use hardware security keys
+
+Physical security keys (like YubiKey) provide strong 2FA protection and are generally much more secure than using an authenticator app or SMS.
```
jviide: It might cause less friction if we focused less on the hardware security keys and instead recommend WebAuthn options that they might already have, be it via their operating systems (Windows, Android, Apple systems, ...) and/or password managers. They're still great against e.g. phishing when compared to TOTP or SMS based options. Separate hardware security keys could be mentioned just as one additional way to use WebAuthn for those that have the means to get them.

Both NPM and GitHub support WebAuthn, NPM just calls them "security keys". GitHub supports WebAuthn passkeys as an full replacement for password + 2FA based logins, but also supports hardware keys as secondary factors.

docs/docs/publishing.md
```diff
@@ -0,0 +1,211 @@
...
+It is possible to specify an `environment` in your workflow:
+
+```yaml [publish.yml]
+jobs:
+  publish:
+    environment: production
+```
+
+In GitHub, you can then configure this environment to require manual approval before the job can proceed. This ensures that even if you manage to trigger the workflow, a human still needs to review and approve the job before it can publish.
+
```
jviide, line 215: When using Trusted Publishing, it's also important to remember to configure the NPM side to require that the publishing job is run in that specific environment.

Image

Otherwise if an attacker gains the ability to push stuff to a repo (like what happened with some of the recent NPM package compromises) they can just push a modified workflow with the same name to a unprotected branch and publish from there using Trusted Publishing. I think that would be a good reason to move this recommendation up higher into the main section of this post.

Of course this isn't perfect either: if the attacker gains a widely scoped GitHub token of yours too, then they can do the approvals themselves (I think?). It's possible to avoid keeping widely scoped GitHub tokens laying around, but hopefully GitHub will implement some sort of additional 2FA requirements for approvals in the future. Fingers crossed!

docs/docs/publishing.md
```diff
@@ -0,0 +1,211 @@
...
+There's currently a feature request and [ongoing discussion](https://github.com/orgs/community/discussions/174507#discussioncomment-14559845) about supporting 2FA for workflow approvals on GitHub.
+
+Until this is supported, using trusted publishing can actually be _less_ secure for solo-maintainers than publishing locally with 2FA enabled.
+
+If your GitHub token leaks somehow, an attack could publish through your trusted workflow the same way a legitimate release would.
+
+Once environment approvals support 2FA, this will no longer be a concern as the attacker would need to pass the 2FA check to approve the workflow.
+
+For these reasons, if you're a solo maintainer, you may want to consider publishing locally with 2FA enabled until this feature is available.
```
jviide: Maybe worth mentioning that other advice that's not directly related to Trusted Publishing, like branch protection, switching phishing resistant 2FA methods etc. still apply.

docs/docs/publishing.md
43081j: they do, actions-up is extremely useful for switching to sha based actions for example

the two are not equivalent, but i agree its worth making this clear

docs/docs/publishing.md
43081j: the intention was to mention this, i just missed it 👍

docs/docs/publishing.md
```diff
@@ -0,0 +1,225 @@
...
+- ✓ Dismiss stale pull request approvals when new commits are pushed
+- ✓ Require approval of the most recent reviewable push
+
+This ensures that nobody can push directly to the `main` branch, and that all changes are reviewed before being merged.
+
+#### Secrets & Variables
+
+Navigate to `Settings > Secrets & Variables > Actions`:
+
+- Remove any existing npm secrets from repository secrets.
```
sxzz: `Remove all unnecessary secrets (if any), such as npm tokens and codecov.io tokens, and use OIDC instead.`

docs/docs/publishing.md
```diff
@@ -0,0 +1,225 @@
...
+Now we need to configure trusted publishing for our npm package. This will ensure that only workflows we trust can publish new versions of our package.
+
+Navigate to your npm package page, and click on the "Settings" tab. You will now see a "Trusted Publishing" section.
+
+In this "Trusted Publishing" section, setup a trusted publisher for the workflow you're about to create with the following settings:
+
+- Organisation and repository are the ones you're publishing from
+- Workflow filename is the name of the workflow file you'll create (e.g. `publish.yml`)
+
+While you're in there, also check the box for "Require two-factor authentication and disallow tokens (recommended)". This will ensure that manual publishing _must_ use 2FA.
```
sxzz, line 73: It is recommended to introduce https://github.com/sxzz/userscripts/blob/main/src/npm-trusted-publisher.md and https://github.com/antfu/open-packages-on-npm until the official npm UI for OIDC is improved.

docs/docs/publishing.md
```diff
@@ -0,0 +1,225 @@
...
+
+- Test (runs tests)
+- Build (builds the package)
+- Publish (publishes the already built package)
+
+There are a few important things to note about this workflow:
+
+- The package is built in a separate job to publishing, ensuring the publish permissions are not exposed to runtime build code
+- All actions are pinned to a full-length commit SHA
+- Install is done with `--ignore-scripts` to avoid running any lifecycle scripts that may be malicious
```
sxzz, line 96: This option may be available only for npm (see pnpm: https://github.com/pnpm/pnpm/issues/3063). We should make a note of this for pnpm users.

docs/docs/publishing.md
```diff
@@ -0,0 +1,225 @@
...
+
+- You must still push tags manually
+- Any time a tag is pushed, a GitHub release is created with a changelog generated from commit messages
+- Any time a tag is pushed, the package is published
+
+## Maintenance & Tooling
+
+### Managing dependency updates
+
+We also highly recommend setting up [dependabot](https://docs.github.com/en/code-security/dependabot) or [renovate](https://www.whitesourcesoftware.com/free-developer-tools/renovate/) to keep your dependencies up to date.
```
sxzz, line 156: I believe it's important to introduce both dependabot/renovate and taze/actions-up. The former is typically deployed on cloud platforms, while the latter is a local tool.

docs/docs/publishing.md
```diff
@@ -0,0 +1,225 @@
...
+```
+
+In GitHub, you can then configure this environment to require manual approval before the job can proceed. This ensures that even if you manage to trigger the workflow, a human still needs to review and approve the job before it can publish.
+
+> [!IMPORTANT]
+> If you do this, ensure that you set the environment in your npm trusted publishing settings too.
+
+### Use hardware security keys
+
+Physical security keys (like YubiKey) provide strong 2FA protection and are generally much more secure than using an authenticator app or SMS.
```
sxzz, line 221: npm is transitioning 2FA from TOTP to WebAuthn.
> The new TOTP (Time-based One-Time Password) setup for npm access will be permanently disabled.
https://github.blog/changelog/2025-09-29-strengthening-npm-security-important-changes-to-authentication-and-token-management/#what-you-need-to-do

I believe this information should be placed under Prerequisites rather than Further Security.

sxzz: However, physical security keys do provide an extra layer of security. Perhaps we could add a recommendation about using WebAuthn (for both GitHub and npm) in the Prerequisites section, and keep this part here.

docs/docs/publishing.md
43081j: this flag does work for pnpm users. the issue is regarding setting it in a `.npmrc` and respecting that afaict (which this isn't doing)

docs/docs/publishing.md
43081j: we do mention actions-up in the next section ("Keeping actions up to date")

though i'm not sure if we should be recommending manual local updates for dependencies. this is meant to be a best practice guide, not an exhaustive list of tools. best practice is not to update your dependencies manually IMO - it is to use an automation and review it in PRs

PR #135, opened 2025-09-19
docs: add `axios` replacements

outslept:
https://github.com/es-tooling/module-replacements/blob/main/docs/modules/axios.md

### DIFF
--- a/docs/guide/replacement-guides/axios.md
+++ b/docs/guide/replacement-guides/axios.md
@@ -11,80 +11,48 @@ The native [`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API)
 Example:
 
 ```js
-import axios from 'axios' // [!code --]
+// GET
+const res = await fetch('https://api.example.com/data')
+const data = await res.json()
 
-const { data } = await axios.get('https://api.example.com/data') // [!code --]
-
-const res = await fetch('https://api.example.com/data') // [!code ++]
-const data = await res.json() // [!code ++]
-
-await axios.post('https://api.example.com/data', { key: 'value' }) // [!code --]
-
-await fetch('https://api.example.com/data', { // [!code ++]
-  method: 'POST', // [!code ++]
-  headers: { 'Content-Type': 'application/json' }, // [!code ++]
-  body: JSON.stringify({ key: 'value' }) // [!code ++]
-}) // [!code ++]
+// POST
+await fetch('https://api.example.com/data', {
+  method: 'POST',
+  headers: { 'Content-Type': 'application/json' },
+  body: JSON.stringify({ key: 'value' })
+})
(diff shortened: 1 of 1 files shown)


docs/guide/replacement-guides/axios.md
```diff
@@ -0,0 +1,90 @@
...
+```
+
+## `ky`
+
+[`ky`](https://github.com/sindresorhus/ky) is a tiny, elegant HTTP client built on top of `fetch`. It offers a nicer API than raw fetch and supports hooks (interceptors), timeouts, and retries.
+
+Example:
+
+```js
+import axios from 'axios' // [!code --]
```
43081j: this isn't really a useful diff since the entire code changes

we should probably just show it as a snippet in that case, i.e. no diff lines

docs/guide/replacement-guides/axios.md
```diff
@@ -0,0 +1,90 @@
...
+const data = await api.get('users').json() // [!code ++]
+```
+
+## `ofetch`
+
+[`ofetch`](https://github.com/unjs/ofetch) is a higher-level fetch wrapper with automatic JSON parsing, request/response interceptors, retries, and a simple API.
+
+Example:
+
+```js
```
43081j, line 49: same for this one, probably drop the diff

docs/guide/replacement-guides/axios.md
```diff
@@ -0,0 +1,90 @@
...
+await fetch('https://api.example.com/data', { // [!code ++]
+  method: 'POST', // [!code ++]
+  headers: { 'Content-Type': 'application/json' }, // [!code ++]
+  body: JSON.stringify({ key: 'value' }) // [!code ++]
+}) // [!code ++]
+```
+
+## `ky`
+
+[`ky`](https://github.com/sindresorhus/ky) is a tiny, elegant HTTP client built on top of `fetch`. It offers a nicer API than raw fetch and supports hooks (interceptors), timeouts, and retries.
```
43081j: let's get rid of the marketing speak ("elegant HTTP client") - its a tiny HTTP client built on top of `fetch` (makes it less opinionated)

does it offer a nicer API than fetch? that too is an opinion i think. we're better saying it adds common helper methods around fetch i think

much of this is what the repo says to try sell itself, but we don't need to do that here. facts only! :D

outslept: Repharased that 👍🏼

docs/guide/replacement-guides/axios.md
outslept: 👍🏼

docs/guide/replacement-guides/axios.md
outslept: 👍🏼

PR #137, opened 2025-09-27
docs: add `fs-extra` replacements

outslept:
https://github.com/es-tooling/module-replacements/blob/main/docs/modules/fs-extra.md

I listed all fs-extra methods in the mapping table (I didn't like other variations for some reason..). This isn’t strictly necessary for sure; it was the simplest initial pass, and I’m happy to trim it down to just the non-1:1 entries (or other variation).

Also I’m aware Deno and other runtimes have their own fs APIs; I intentionally didn’t include non-Node guidance. The expectation is that non‑Node users will rely on a Node compatibility layer of their tooling.


43081j:
this seems good but i do wonder if we should just have a subheading for each one (e.g. to show what you mean by things such as `Consider maxRetries/retryDelay.`).

ill merge this for now but we may revisit it

PR #156, opened 2026-01-29
docs: add gray-matter replacement

remcohaszing:
I wrote a tiny frontmatter parser. I felt the need, because existing parsers have outdated dependencies.

Of course `simplematter` isn’t as battle-tested as `gray-matter`. I also kept it minimal on purpose. It doesn’t support specifying the language on a frontmatter fence, custom delimiters, or reading files.

### DIFF
--- a/docs/docs/replacements/gray-matter.md
+++ b/docs/docs/replacements/gray-matter.md
@@ -1,0 +1,29 @@
+---
+description: Modern frontmatter parsers
+---
+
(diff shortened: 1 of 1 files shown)


43081j:
what are you currently using it in? it'd be good to see some real world usage before we merge this

i don't want to rely on download count but also don't want to add new things too soon. so even just references to any projects using it today would be helpful (here in the issue, not in the docs)

PR #160, opened 2026-02-02
docs: list npmx in resources/websites

jhroemer:
Hey there 👋 

Love the learn section and use it often 🙇  Would be really cool to list [npmx](https://npmx.dev/) too, so that more people can benefit from it. @danielroe not sure if you're ok with spreading the word even more at this point. I'm not entirely sure about the copy, figured rather keep it concise.

### DIFF
--- a/docs/learn/resources.md
+++ b/docs/learn/resources.md
@@ -42,7 +42,7 @@ description: Tools, websites, and resources for improving the JavaScript ecosyst
 | [pkg-size.dev](https://pkg-size.dev/) | Explores npm packages with fresh installs to show actual size, dependencies, and helps optimize your node_modules |
 | [snyk](https://snyk.io/) | Finds and fixes vulnerabilities in open source dependencies |
 | [socket.dev](https://socket.dev/) | Security-focused tool that analyzes dependencies for vulnerabilities |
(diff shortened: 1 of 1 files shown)


docs/learn/resources.md
```diff
@@ -42,6 +42,7 @@ description: Tools, websites, and resources for improving the JavaScript ecosyst
 | [pkg-size.dev](https://pkg-size.dev/) | Explores npm packages with fresh installs to show actual size, dependencies, and helps optimize your node_modules |
 | [snyk](https://snyk.io/) | Finds and fixes vulnerabilities in open source dependencies |
 | [socket.dev](https://socket.dev/) | Security-focused tool that analyzes dependencies for vulnerabilities |
+| [npmx.dev](https://npmx.dev/) | A better browser for the npm registry |
```
43081j: given this is on the e18e site, we should probably mention how it relates - specifically, npmx shows e18e recommendations.

so maybe something like:

> A better browser for the npm registry with e18e recommendations and advice

or

> A better browser for the npm registry which also shows e18e advice and recommendations

jhroemer: Yes, that's great! Thanks! I personally prefer the former (which I pushed).

43081j (approved):

PR #168, opened 2026-02-12
feat: add sync sponsors workflow

ghostdevv:
Basically a copy of the workflow [sync-replacements](https://github.com/e18e/e18e/blob/ddf9d61446d9ceffdea4f1bc7aa0e02f8123eb4d/.github/workflows/sync-replacements.yml) workflow with the sponsorkit changes instead

### DIFF
--- a/.github/workflows/sync-sponsors.yml
+++ b/.github/workflows/sync-sponsors.yml
@@ -1,0 +1,49 @@
+name: Sync module replacements
+
+on:
+  schedule:
+    - cron: '0 9 * * 1' # Weekly on Monday at 9am UTC
+  workflow_dispatch:
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  sync:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


ghostdevv:
I also wonder whether it should push to main instead of opening a PR? that would make it less annoying to run it daily for example.

.github/workflows/sync-sponsors.yml
```diff
@@ -0,0 +1,49 @@
...
+
+      - name: Create pull request
+        if: steps.changes.outputs.changed == 'true'
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git checkout -b sync-sponsors
+          git add scripts/module-replacements.sha docs/docs/replacements/index.md
```
43081j: i guess this bit should add the image instead?

ghostdevv: ah yea, missed that xD

PR #177, opened 2026-03-03
docs: add sponsors page

ghostdevv:
Add a sponsorship page so we can link to it from places off-site, like the sponsor image cards in the readme, so that we can update that page as needed without having to update individual places every time


43081j (approved):
good idea!

looks good to me 👍

PR #196, opened 2026-04-02
docs(cli): rework cli docs and point readers to the repo

dreyfus92:


ghostdevv (changes requested):
could we add redirects to the new pages from the old ones 👀

43081j (approved):

PR #197, opened 2026-04-04
docs: render sponsor links as HTML for better discoverability

shtefcs:
Right now all sponsors on the homepage are displayed as a single SVG image via ``. Browsers treat this as an opaque image, so the links inside the SVG are invisible to search engines and screen readers. None of the sponsor websites are actually linked in the page HTML.

I think sponsors should get a proper, clickable link back to their website. Right now they get a logo inside an image and nothing else. If sponsors can see their contribution is properly credited with a real link from e18e.dev, that makes the sponsorship more valuable and could attract more support.

I opened similar PRs for [UnoCSS](https://github.com/unocss/unocss/pull/5165) and [VueUse](https://github.com/vueuse/vueuse/pull/5348) which had the same problem with the `antfu/static` sponsors SVG.

**What this does:**

- Adds an `onSponsorsReady` callback to `sponsor.config.ts` that writes `docs/public/sponsors.json` alongside the existing SVG. This gives the website structured sponsor data with proper tier classification and links.
- Adds a VitePress data loader that reads the local JSON (zero API calls at build time) and enriches URLs from OpenCollective profiles when sponsors only link to their OC page.
- Creates a `SponsorLinks.vue` component that renders real `<a>` links with circular avatars, grouped by tier (Platinum, Gold, Silver, Sponsors, Backers).
- All links are dofollow (`rel="noopener"` only).
- Handles `Infinity` monthly amounts (Platinum sponsors like Chrome) by serializing them as `999999` in JSON.

**One small manual step needed:** the `sync-sponsors.yml` workflow should also `git add docs/public/sponsors.json` alongside `docs/public/sponsors.svg` so the JSON gets committed on each weekly sync. I could not include the workflow change in this PR due to GitHub token permissions, but the diff is just changing line 45 from:
```
git add docs/public/sponsors.svg
```
to:
```
git add docs/public/sponsors.svg docs/public/sponsors.json
```

**Note:** The included `sponsors.json` is a placeholder generated from the public OpenCollective API. Once the sponsorkit workflow runs with the API key, it will produce accurate monthly amounts and proper tier classification.

No new dependencies.

### DIFF
--- a/docs/.vitepress/theme/components/SponsorLinks.vue
+++ b/docs/.vitepress/theme/components/SponsorLinks.vue
@@ -1,0 +1,107 @@
+<script setup lang="ts">
+import { data as sponsors } from '../sponsors.data'
+</script>
+
+<template>
+  <div class="sponsor-links">
+    <div v-for="tier in sponsors" :key="tier.tier" class="sponsor-tier">
+      <h4 class="sponsor-tier-title">
+        {{ tier.tier }}
+      </h4>
+      <div class="sponsor-grid" :class="`sponsor-grid-${tier.size}`">
+        <a
+          v-for="item in tier.items"
+          :key="item.url"
+          :href="item.url"
+          target="_blank"
+          rel="noopener"
+          :title="item.name"
+          class="sponsor-link"
+        >
+          <img
+            :src="item.img"
+            :alt="item.name"
+            class="sponsor-avatar"
+            loading="lazy"
+          >
+        </a>
+      </div>
+    </div>
+  </div>
+</template>
+
+<style scoped>
+.sponsor-links {
+  max-width: 768px;
+  margin: 1.5rem auto 0;
+}
+
+.sponsor-tier {
+  margin-bottom: 2rem;
+}
+
+.sponsor-tier-title {
+  text-align: center;
+  font-size: 0.875rem;
+  font-weight: 400;
+  opacity: 0.5;
+  margin-bottom: 1rem;
+}
+
+.sponsor-grid {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: center;
+  align-items: center;
+}
+
+.sponsor-link {
+  display: block;
+  transition: opacity 0.2s;
+  opacity: 0.8;
+}
+
+.sponsor-link:hover {
+  opacity: 1;
+}
+
+.sponsor-avatar {
+  border-radius: 50%;
+}
+
+.sponsor-grid-big {
+  gap: 1rem;
+}
+
+.sponsor-grid-big .sponsor-avatar {
+  width: 80px;
+  height: 80px;
+}
+
+.sponsor-grid-medium {
+  gap: 0.75rem;
+}
+
+.sponsor-grid-medium .sponsor-avatar {
+  width: 64px;
+  height: 64px;
+}
+
+.sponsor-grid-small {
+  gap: 0.625rem;
+}
+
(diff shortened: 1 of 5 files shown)


bluwy:
You can use an object tag to render the svg so you can click in it. I've used this in my website:

https://github.com/bluwy/website/blob/039957480daff6bab778d07e6ad67303de3264bd/src/routes/%2Bpage.svelte#L108-L114

https://bjornlu.com/#:~:text=Sponsors

shtefcs:
Thanks for the suggestion! Using `<object>` would definitely make the SVG links clickable for users, which is a nice improvement over ``.

The main goal here though is to get the sponsor links picked up by search engines so sponsors actually get backlink credit. Links inside an `<object>` tag live in a separate browsing context (similar to an iframe), and search engines don't follow those links for ranking purposes. So while users could click them, Google wouldn't see them as part of the page.

That's why I went with rendering real `<a>` tags directly in the page HTML. It's the only way to make sure sponsors get proper link value from being listed on e18e.dev.

ghostdevv:
> The main goal here though is to get the sponsor links picked up by search engines so sponsors actually get backlink credit.

Personally I'd rather not end up in a [socket.io](https://socket.io/) situation, so I'd be up for Bluwy's suggestion

shtefcs:
That's a fair point. I'm aware of the casino/gambling sponsor issue that hit socket.io and other projects, and I definitely don't want to open the door to that here.

The real problem there was that anyone could pay a small amount on Open Collective and automatically get a dofollow backlink with zero review. That's what made those projects easy targets.

I think the solution is manual curation. Instead of pulling directly from the Open Collective API, the project could maintain a simple allowlist (a JSON file or array in the config) of approved sponsor logins. Only sponsors on that list would get rendered as links on the site. New sponsors would still show up in the SVG as before, but they wouldn't get a link until someone on the team reviews and approves them.

That way legitimate sponsors from the ecosystem still get properly credited, but random casino accounts can't buy their way into a dofollow backlink just by dropping a few dollars on Open Collective.

I can update the PR to work this way if you're open to it. Happy to make it easy to maintain too, something like an `approved-sponsors.json` that the team controls.

ghostdevv:
I think where I was going was to keep the behavior of no backlinks, while still allowing people to click through by using bluwy's suggestion of `<object>`. I _think_ it's normal to add noreferrer/nofollow to prevent backlinks, but I could be wrong here.

bluwy:
Yeah I don't think we necessarily need the backlinks too. A clickable link should be enough. The folks who sponsor would believe in the project itself than needing the backlink credit.

43081j:
just catching up on this, sorry it took me a while.

i think the way forward here is to do what @bluwy said and ensure the SVG itself has real links etc.

i understand having backlinks is beneficial, but this does greatly increase the complexity of the sponsors generation and introduces the risk of sponsorships being more for the search engine gains than for good intent. it'd be nice to keep the latter especially - sponsorships happening because of belief in the initiative and/or collaborations rather than for exposure alone.

in projects where there is a "product" (like SaaS), i'd probably lean more into this change since the main reason of sponsorship would then be such exposure. in our case, its a community effort being sponsored so the goals are different.

i don't want to lose potential sponsorships for the community but feel like most or all currently won't have a concern about this at least.

hopefully this makes sense. if not, please do come discuss on our discord or something as it would be nice to understand better.

if someone can open a PR for the SVG change bjorn suggested, that would be much appreciated 🙏

PR #208, opened 2026-05-15
chore: fix sync-replacements workflow

ghostdevv:
We can pretend I didn't break it in #205

### DIFF
--- a/.github/workflows/sync-replacements.yml
+++ b/.github/workflows/sync-replacements.yml
@@ -47,11 +47,15 @@ jobs:
           node-version: 22
 
       - name: Use pnpm
-        uses: pnpm/action-setup@41ff72655975bd51cab0327fa583b6e92b6d3061 # v4.2.0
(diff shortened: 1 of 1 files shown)


dreyfus92 (approved):
looks good 😄

gameroman (approved):

43081j:
How come these are not the defaults? 

i.e what is working directory by default? And the package path by default? 

They're not the root directory and the package.json in it? 🤔

Would be good to understand what broke it

ghostdevv:
> How come these are not the defaults?
> 
> i.e what is working directory by default? And the package path by default?
> 
> They're not the root directory and the package.json in it? 🤔
> 
> Would be good to understand what broke it

It's because we're cloning the e18e repo into the `e18e` folder instead of the root (which is some directory that github creates) because we need to also clone the `module-replacements` repo. When I opened #205 I clearly didn't test it because I would've seen that I forgot to run these commands in the `e18e` sub directory.

e.g. all action usually run from `/foo` but we're cloning to `/foo/e18e` (and `/foo/module-replacements`) so we need to run the commands from `/foo/e18e`

PR #210, opened 2026-05-21
chore: add depot to sponsors as infra partner

ghostdevv:
Added Depot to the sponsors image as an Infra Partner :partying_face: 

This seems to be the least annoying way to do it, but it'll break if someone somehow also donates `49.99` but we can cross that bridge if we come to it xD


dreyfus92 (approved):

AlexanderKaran:
Thank you @ghostdevv

43081j (approved):

PR #222, opened 2026-06-24
feat: add vitepress llms.txt plugin

### LINKED ISSUE #223: Provide `llms.txt` for Vitepress website
Generates LLMs friendly (markdown) versions of the website's pages, according to the [llms.txt](https://llmstxt.org) spec.

It can be done using the [`vitepress-plugin-llms`](https://github.com/okineadev/vitepress-plugin-llms) plugin which is notably used on Vitepress own website.

AFCMS:
Generates pages as LLMs friendly markdown files at build time (from the [llms.txt](https://llmstxt.org) spec)

It uses the [`vitepress-plugin-llms`](https://github.com/okineadev/vitepress-plugin-llms) plugin which is notably used on Vitepress own website.

Output files:
- `llms.txt`
- `llms-full.txt`
- A markdown version of each page

Fixes #223

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -2,7 +2,7 @@ import process from 'node:process'
 import { transformerTwoslash } from '@shikijs/vitepress-twoslash'
 import { defineConfig } from 'vitepress'
 import { groupIconMdPlugin, groupIconVitePlugin } from 'vitepress-plugin-group-icons'
-import llmstxt from 'vitepress-plugin-llms'
+import llmstxt, { copyOrDownloadAsMarkdownButtons } from 'vitepress-plugin-llms'
 import { buildEnd } from './buildEnd.config'
 
 const ogTitle = 'e18e'
@@ -170,6 +170,7 @@ export default defineConfig({
     codeTransformers: [transformerTwoslash()],
     config(md) {
       md.use(groupIconMdPlugin)
+      md.use(copyOrDownloadAsMarkdownButtons)
     },
   },
   vite: {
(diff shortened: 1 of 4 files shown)


gameroman (commented):
I think it would be better to open an issue for discussion first

AFCMS:
> I think it would be better to open an issue for discussion first

Done.

ghostdevv (approved):

gameroman (commented):
Would be nice if there was also a "Copy as Markdown" button

gameroman (approved):
I like this

dreyfus92 (approved):
thanks everybody ✌🏻

gameroman:
cc @43081j

gameroman (changes requested):
@AFCMS I see it several nice features

Image

But I think we only need `Copy as Markdown` on the page, we probably don't need "Open in ChatGPT/Claude" or a download button

43081j:
change looks good to me 👍 

ill wait until we get an answer to @gameroman's question though. if it isn't configurable, its probably fine as it is.

AFCMS:
> But I think we only need `Copy as Markdown` on the page, we probably don't need "Open in ChatGPT/Claude" or a download button

The button provided by the plugin isn't configurable, sadly. If needed I can try to make custom one.

gameroman:
> > But I think we only need `Copy as Markdown` on the page, we probably don't need "Open in ChatGPT/Claude" or a download button
> 
> The button provided by the plugin isn't configurable, sadly. If needed I can try to make custom one.

It's fine for now then 👍

gameroman (approved):

ghostdevv:
On second look I wouldn't mind a custom button so it's less big and maybe placed better, like alongside the header or in the sidebar. However, idm if we look at that in a follow up

43081j:
let's take a look at it in a follow-up 👍

okineadev:
> > But I think we only need `Copy as Markdown` on the page, we probably don't need "Open in ChatGPT/Claude" or a download button
> 
> The button provided by the plugin isn't configurable, sadly. If needed I can try to make custom one.

I'll take a look at this soon.

Thanks for this PR too!

PR #236, opened 2026-07-29
docs: add sourcemaps blog post

43081j:


TheAlexLichter:
* We should explicitly mention that is is for *libraries* only. This can be implicitly inferred from the context and is mentioned later on (" On the other hand, if you're building an app, you should absolutely do these things to reduce the size of your runtime.") but I'd say making it clear earlier would be better.
* OG is still the old one (obvious but still wanted to point that out)

Otherwise good primer on source maps (TODO chapter pending that'll also help a lot).

TheAlexLichter (approved):

PR #240, opened 2026-08-04
docs: add note about biome's noRestrictedDependencies

ghostdevv:
Document Biome's new(ish) module-replacements rule https://biomejs.dev/linter/rules/no-restricted-dependencies/

### DIFF
--- a/docs/docs/replacements/index.md
+++ b/docs/docs/replacements/index.md
@@ -1,6 +1,6 @@
 # Replacements
 
-When using the [e18e ESLint plugin](https://github.com/e18e/eslint-plugin) or the [CLI](https://github.com/e18e/cli), you may be suggested replacement modules for dependencies you have. This is a list of replacements, along with their migration guides and any useful information.
+When using the [ESLint plugin](https://github.com/es-tooling/eslint-plugin-depend), the [CLI](https://github.com/e18e/cli), or [Biome's noRestrictedDependencies rule](https://biomejs.dev/linter/rules/no-restricted-dependencies/), you may be suggested replacement modules for dependencies you have. This is a list of replacements, along with their migration guides and any useful information.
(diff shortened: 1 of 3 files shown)


docs/docs/replacements/index.md
```diff
@@ -1,6 +1,6 @@
 # Replacements
 
-When using the [ESLint plugin](https://github.com/es-tooling/eslint-plugin-depend) or the [CLI](https://github.com/e18e/cli), you may be suggested replacement modules for dependencies you have. This is a list of replacements, along with their migration guides and any useful information.
+When using the [ESLint plugin](https://github.com/es-tooling/eslint-plugin-depend), the [CLI](https://github.com/e18e/cli), or [Biome's noRestrictedDependencies rule](https://biomejs.dev/linter/rules/no-restricted-dependencies/), you may be suggested replacement modules for dependencies you have. This is a list of replacements, along with their migration guides and any useful information.
```
bluwy: Not entirely sure if we should add more third-party tooling here, like say if oxlint implements one day, or if npmx also counts, this could get quite long.

bluwy (approved):

docs/docs/replacements/index.md
ghostdevv: I'm ok with that, and it'll be case by case anyway (like I wouldn't add npmx here probably, but it could make sense in a different list somewhere else). People use different tooling and we should help show them that they can keep using it because their tooling actually implements it, or integrates with it.

43081j: i too would discard this part of the PR. this was meant to just list the official routes, knowing there other third party ones we don't exhaustively list

bluwy (approved):

gameroman (approved):

